PR #159, opened 2014-05-09
Compatibility with IE, contenteditables in iframes, and more dynamic activation/deactivation of Scribe

daniel-nelson:
We've made these changes to support our use of Scribe in the [Populr](https://populr.me) editor.

There are a few places where we hard-coded some class names into the span checks, which is a hack to support our specific use case that shouldn't be merged (search for pplr to see what I mean).

All changes are intended to be cross-browser compatible (no browser specific hacks). It has been tested in IE11, and we intend for it to work in IE10.

Anonymous functions were replaced with named functions for event listeners so that they could be removed upon Scribe deactivation. (We were running in a situation in which Scribe was deactivated and reactivated on a particular element, after which "paste" would paste in multiple copies.)

I didn't add any test cases, so I understand that this doesn't meet your requirements for a pull request, but I thought maybe it would be useful to share with you even if it doesn't actually get merged.


rrees:
Closing this, further conversation can happen on #286

PR #175, opened 2014-05-20
Add "browserify" support

TooTallNate:
Hello scribe maintainers! I've added another method of consuming the scribe editor: [browserify](http://browserify.org/)

Changes required:
1. Change `lodash-modern` require calls to `lodash-amd`, because this is the way that npm will install the `lodash-amd` module by default. We can make bower emulate the npm behavior easily.
2. Add the `main` and `dependencies` field to the `package.json` file so that browserify knows the JavaScript entry point and npm knows the dependencies to install.
3. Add the `browserify.transform` field to the `package.json` so that browserify knows to "deamdify" the scribe code into CommonJS.
4. Updated the `Plumbing.js` build file to match the npm install output.

So now users can consume scribe via browserify via CommonJS code like this:

``` js
var Scribe = require('scribe');

var scribe = new Scribe(document.querySelector('rte'));
scribe.setContent('<p>Hello, World!</p>');

scribe.on('content-changed', updateHtml);
function updateHtml() {
  document.querySelector('.rte-output').value = scribe.getHTML();
}

updateHtml();
```

<del>__Note:__ depends on https://github.com/lodash/lodash-amd/pull/2 (and subsequently a new `lodash-amd` version update) in order to work properly.</del>

Any shot of this being merged into master branch? Cheers!

### DIFF
--- a/.tern-project
+++ b/.tern-project
@@ -4,7 +4,7 @@
       "baseURL": "./src",
       "paths": {
         "html-janitor": "./bower_components/html-janitor/src/html-janitor",
-        "lodash-modern": "../bower_components/lodash-amd/modern"
+        "lodash-amd": "../bower_components/lodash-amd"
       }
     },
     "node": {}


ishaikovsky:
Just FYI (I see it's been a long time but still). 

Global "deamdify" didn't work for us because there are also other good libraries but with bad support of such environment in our project. E.g. in our setup - Underscore is starting to fail with undefined during initialization when global "deamdify" transform invoked. And works fine if not. I found couple more reports of the same and finally found a bug report that was submitted to them confirmed and merged to master.... but never released. https://github.com/jashkenas/underscore/commit/7fc15e5816dc02371517c4cb16a28e35f5b9a857.

So my personal IMO - would have been better to localize your requirement to run "deamdify" against just Scribe modules (even by repeating this few lines in each package..) rather than forcing projects to use global option.

PR #213, opened 2014-07-01
Smart formatters 

bliker:
I started digging around the concept and this is what I have got so far.

First commit pretty much rips out all the previous formatting code. As it does not fit with my notion of formatters that are now just simple plugins. They register their own mutation observers and do manipulations based on that.

For managing mutations am using [rafaelw/mutation-summary](https://github.com/rafaelw/mutation-summary) from google as a helper, but the useful functionality can be ripped out in future, as it is big and without AMD support (that is why my repo is in bower).

45c71b2 removes everything about undo manager. Because it is no longer necessary.

**This is still work in progress**


rrees:
I think we'd need this change resubmitted based on a new merge with upstream

PR #235, opened 2014-07-23
CommonJS distribution

callum:
The app I'm working on has a plugin architecture like Scribe. One of the plugins depends on Scribe, but it's proven quite difficult to manage the dependencies between CommonJS- and AMD-based modules.

Based off of @TooTallNate's CommonJS example I'm proposing this solution. The idea is that this script would be run when a release happens. It uses Browserify's standalone option (https://github.com/substack/browserify-handbook#standalone).

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "scribe-editor",
   "version": "0.1.17",
-  "main": "src/scribe.js",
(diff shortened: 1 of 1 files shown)


rrees:
We've moved on with the discussion around ES6 modules, if you're still interested please discuss on #314

PR #242, opened 2014-08-12
Issues with command state/value overrides 

csinchok:
So, I've noticed that in the example editor, hitting bold, italic, etc before typing doesn't always work as expected.

If you type a few non-space characters, and hit command-B, the *next* character you type will be bold. However, if you type a space character (translated to an `&nbsp;`), and hit command-B, it appears that nothing happens.

After some investigation, I believe that scribe is inadvertently causing the [contenteditable state/value](https://dvcs.w3.org/hg/editing/raw-file/tip/editing.html#state-override) overrides to get reset, probably while running formatters.

I've written a small test to demonstrate this, and have temporarily patched the issue in command-patch.js, by just not running the transaction manager on an empty selection. Perhaps there's a better place to make this change? Would love some feedback on this.

### DIFF
--- a/src/api/command-patch.js
+++ b/src/api/command-patch.js
@@ -8,9 +8,13 @@ define(function () {
     }
 
     CommandPatch.prototype.execute = function (value) {
-      scribe.transactionManager.run(function () {
+      if (scribe.selection === undefined) {
         document.execCommand(this.commandName, false, value || null);
-      }.bind(this));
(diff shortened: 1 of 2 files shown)


rrees:
@csinchok if you're still interested in this can you sync with master and see if the tests pass?

rrees:
Closing for now

PR #250, opened 2014-08-30
Is the bold command patch really necessary?

csinchok:
So, currently, the bold command patch prevents a user from running the bold command inside of a heading tag. A comment in that patch seems to be referencing [this example](http://jsbin.com/IxiSeYO/4/edit).

The problematic bit in this patch happens when a user first makes text bold, and *then* decides to make it a heading. In this case, you end up with `<h2><b>Some bolded heading</b></h2>`, and then the user isn't able to "de-bold" the heading.

As an experiment, I made a branch that removed this command patch and...everything seemed to function properly. While I still wasn't able to bold text inside a heading, I *was* able to remove the bold tag from a heading like this: `<h2><b>Some bolded heading</b></h2>`. Also, I didn't see any bad markup in Chrome, Safari or Firefox.

Any guidance on this would be greatly appreciated. Is there something that I'm missing here?


rrees:
@csinchok Does the changes in 1.2.0 allow you to override the behaviour now?

rrees:
I'm closing this because I think we've resolved the issue

Issue #95, opened 2014-04-09
Unbind/destroy


OliverJAsh:
When destroying a Scribe instance, it should clean-up all references and DOM event bindings, including in any plugins being used.

saranrapjs:
+1 — currently working around this by cloning elements with DOM listeners (like the toolbar) but would be great if "destroy" was baked into the plugin architecture somehow

anatoliyarkhipov:
+1

triptec:
Any news about this? I'm also putting this in a react app and would like to clean up.

rrees:
I think we're looking for a suitable community contribution.

triptec:
@rrees is there something simple to start with, like adding it to the api and set this = null and then step by step clean up plugins, event listeners et al?

rrees:
@triptec I love changes that are simple so this sounds like a good place to start.

triptec:
I'll look into it this weekend.

rrees:
Thanks feel free to just implement the basic stub and then add the event listeners and plugins in a follow up.

PR #288, opened 2014-11-04
Unbind DOM events on "destroy" event

craigspaeth:
This isn't merge-ready yet, but I wanted to get the ball rolling on a PR for https://github.com/guardian/scribe/issues/95.

This unbinds DOM events when calling `scribe.destroy()`. I imagine there's more to tear down so I was wondering if anyone might have ideas on what else needs to happen in destroy.

Also this is obviously missing tests. I'm having a little trouble figuring out how to test this as all the other tests seem like high-level integration tests that test the output HTML. It [doesn't seem like I can inspect the attached event handlers](http://stackoverflow.com/questions/2623118/inspect-attached-event-handlers-for-any-dom-element) from the outside, so I'm not sure how to test if these event handlers have been removed. However, I'm happy to dig for a testing solution, just wondered if anyone has insight on this first.

### DIFF
--- a/src/scribe.js
+++ b/src/scribe.js
@@ -248,6 +248,11 @@ define([
     this._plainTextFormatterFactory.formatters.push(fn);
   };
 
+  Scribe.prototype.destroy = function (configurePlugin) {
+    this.trigger('destroy');
+    return this;
+  };
+
   // TODO: abstract
   function FormatterFactory() {
     this.formatters = [];


mindplay-dk:
I hacked this in myself, without altering the codebase.

```
    Scribe.prototype.destroy = function () {
        this.trigger("destroy");
    };
```

Of course this doesn't address memory issues created by core plugins that aren't actually using this event to clean up, but I needed this event to help clean up residual UI generated by some of my own plugins when their editors are destroyed.

This PR has been sitting still for far too long - this really needs to get into the main codebase.

Even if it's hard to test whether the individual plugins get cleaned up properly, this feature is immediately useful for cleaning up UI artifacts generated by plugins.

rrees:
Okay, I think we can add the event in a separate PR.

The rest of it needs to be introduced in a more considered way.

mindplay-dk:
> The rest of it needs to be introduced in a more considered way.

Considered?

Looks like @craigspaeth has already done most of the work of adding the missing event clean-ups - isn't that the (only) point of adding this event?

Likely all internal plug-ins need a full review though, before merging this, to see if everything is still getting cleaned up, as this fork is quite old by now - likely new event handlers have been added since then...

PR #290, opened 2014-11-07
Add support for iframes

EmilS:
This PR builds off of https://github.com/guardian/scribe/pull/221

First change is to have all document calls use `scribe.targetDocument`. Current API usage: `scribeElement` passed in is a div inside of an iframe and the options object has `iframe` set to true, e.g. 

```javascript
// scribeElement is an element inside an iframe
// Create an instance of Scribe
var scribe = new Scribe(scribeElement, { iframe: true });
```

I'm looking for opinions on how I should test the changes. I considered running the testsuite twice (once with the div inside an iframe, once not), but that may be going overboard.

@OliverJAsh @hmgibson23


EmilS:
@hmgibson23 I've setup a [jsfiddle](http://jsfiddle.net/9f7qme9e/10/) running the latest Scribe code on my branch. The iframe is visible, but works in Firefox.

hmgibson23:
Okay - well it looks like the build is passing now anyway. I'm happy for this to be merged. 

:+1: 


\cc @rrees for a second pair of eyes

rrees:
Looks okay to me although I feel we should document the use of `scribe.targetWindow` and `scribe.targetDocument`

hmgibson23:
Created an issue for the documentation and will merge this. https://github.com/guardian/scribe/issues/317

aclever:
@es Any suggestions on making the toolbar work inside the editor as well. We managed to get the content area working, but are struggling with the toolbar, which in our case is also inside an iframe. 

Any thoughts? 

many thanks in advance

PR #295, opened 2014-11-21
Incorporate ImmutableJS

rrees:
OMG! Totes immutes!

This introduces the ImmutableJS library, converts the formatters over to Immutable structures and as a consequence removes the LoDash _flatten_ function.

I also tried to replace _defaults_ with merged maps but there is some subtly in the logic that I am missing.

I also attach the Immutable library to the scribe object so that plugins can make use of it.
## QA

 This doesn't change any functionality, just the way it is implemented. In theory it is a _genuine for-real refactoring_ with passing tests.

However the way the formatters get registered has changed so it might be worth checking the default behaviour of the standard plugins and the external plugins may need to be rechecked as they get deployed.

### DIFF
--- a/src/scribe.js
+++ b/src/scribe.js
@@ -276,10 +276,10 @@ define([
     this.formatters = {
       // Configurable sanitization of the HTML, e.g. converting/filter/removing
       // elements
-      sanitize: Immutable.List([]),
+      sanitize: Immutable.List(),
       // Normalize content to ensure it is ready for interaction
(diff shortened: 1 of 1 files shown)


chrismendis:
Why weren't the command and commandPatches objects refactored in to Immutable Maps?

rrees:
@chrismendis There wasn't a big reason, ultimately we want to do that. Pull request?

PR #299, opened 2014-12-02
Polyfill MutationObserver

nchase:
Currently, MutationObserver is depended upon, but it isn't polyfilled.

This will fail in versions of IE < 11 – what versions IE and other browsers does Scribe officially support?

It will also fail in PhantomJS, which is using a slightly outdated version of Webkit, (MutationObserver support [is planned for PhantomJS 2.0](https://github.com/ariya/phantomjs/issues/10715#issuecomment-50734088), which has not been released yet) so anyone who is using PhantomJS (e.g. running tests on pages that use Scribe via PhantomJS) will currently see errors.

The polyfill used here is cribbed from [megawac/MutationObserver.js](https://github.com/megawac/MutationObserver.js).

### DIFF
--- a/src/dom-observer.js
+++ b/src/dom-observer.js
@@ -25,8 +25,565 @@ define([
       return realChangedNodes.length > 0;
     }
 
-    var MutationObserver = window.MutationObserver || window.WebKitMutationObserver || window.MozMutationObserver;
-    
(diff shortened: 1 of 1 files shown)


rrees:
@nchase I applied the polyfill to our fork of Scribe where we plan to try and have community support for IE. However the test suite fails so this is not the only blocker to supporting IE10.

I'm going to close this but contributions are welcome on the fork.

PR #318, opened 2015-01-14
Revert "Add support for iframes"

hmgibson23:
Reverts guardian/scribe#290


hmgibson23:
I'm reverting this because of the timing issues with builds that we saw. As we don't need to use iframes and other features internally, we've spun out a forked version of scribe https://github.com/scribe-editor/scribe to use features like this. If you re-submit this PR there, then I'll merge it in. Sorry about the hassle.

PR #319, opened 2015-01-14
Add support for paste formatters

### LINKED ISSUE #241: Feature request: 'paste' formatter type
Currently sanitizers that use `scribe.registerHTMLFormatter('sanitize', ...)` will sanitize both content that's pasted as well as content that is programmatically added to the editor.

However, I'd like to handle these cases separately. I'd like pasted content to be sanitized to a stricter subset of allowed tags than content that I programmatically inject into the editor. For example, I have code that adds divs to the editor, but I don't want someone to be able to paste them in.

My current workaround is to take over handling the 'paste' event from scribe. I sanitize the content and then place it into the editor at the cursor position.

I'd like to propose a new HTML Formatter type that is called by in Scribe's 'paste' handler that will give client code an opportunity to sanitize only pasted content. Example code would look like:

```
scribe.registerHTMLFormatter('paste', function(html){ return sanitizePastedContent(html); });
```

I can file a PR for this if this proposal is OK with you.

bregenspan:
This is all @awentzonline's work (thank you!!), just with the addition of a test. This should resolve #241 if all looks good.  It depends on https://github.com/guardian/scribe-test-harness/pull/10 being merged and released as it needs a new helper method for inserting via paste (if it's simpler, I could instead inline that for now).

Because the example paste formatter used for the test is kind of a one-off, I wasn't totally sure where to put it -- it lives in "examples" for now.

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
     "q": "~1.0.0",
     "request": "~2.33.0",
     "selenium-webdriver": "~2.41.0",
-    "scribe-test-harness": "~0.0.7"
+    "scribe-test-harness": "~0.0.14"
   },
   "scripts": {
     "test": "./run-tests.sh"
--- a/examples/example-paste-formatter.js
+++ b/examples/example-paste-formatter.js
@@ -1,0 +1,25 @@
+/**
+ * This is a simple paste formatter example which
+ * removes style attributes from all elements that have
+ * them.
+ */
+define(function () {
+
+  'use strict';
+
+  return function () {
+    return function (scribe) {
+
+      scribe.registerHTMLFormatter('paste', function (html) {
+        var bin = document.createElement('div');
+        bin.innerHTML = html;
+        var styledEls = bin.querySelectorAll('[style]');
+        Array.prototype.forEach.call(styledEls, function (el) {
+          el.removeAttribute('style');
+        });
+        return bin.innerHTML;
+      });
+    };
+  };
+
+});
--- a/src/plugins/core/events.js
+++ b/src/plugins/core/events.js
@@ -218,8 +218,9 @@ define([
           event.preventDefault();
 
           if (contains(event.clipboardData.types, 'text/html')) {
-
-            scribe.insertHTML(event.clipboardData.getData('text/html'));
+            var html = event.clipboardData.getData('text/html');
+            html = scribe._htmlFormatterFactory.formatPaste(html);
+            scribe.insertHTML(html);
           } else {
             scribe.insertPlainText(event.clipboardData.getData('text/plain'));
           }
@@ -265,6 +266,7 @@ define([
              */
             scribe.el.focus();
 
+            data = scribe._htmlFormatterFactory.formatPaste(data);
             scribe.insertHTML(data);
           }, 1);
         }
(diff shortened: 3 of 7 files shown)


package.json
```diff
@@ -23,7 +23,7 @@
     "q": "~1.0.0",
     "request": "~2.33.0",
     "selenium-webdriver": "~2.41.0",
-    "scribe-test-harness": "~0.0.7"
+    "scribe-test-harness": "~0.0.14"
```
bregenspan: This is just a placeholder version number as I have a separate PR out for the harness that would need to be released first.

rrees:
Brilliant! It's been in the back of my mind for a while but didn't really have the use case. Let's try and get the test harness released first and then we'll look at the test suite.

package.json
hmgibson23: It's bumped to ```0.0.14``` now.

hmgibson23:
Looks great! :+1:

rrees:
@bregenspan We tried merging but couldn't get the test suite to work on Chrome. No content seems to get pasted to Scribe. We weren't sure how to progress, do you have an idea of what might be wrong?

bregenspan:
@rrees argh, sorry about that! Worked for me locally in Chrome Version 41.0.2272.3 dev on OSX but because this involves focus/selections and there are so many weird Webdriver focus-related issues, I should've checked against Travis first. I might not have time to fix up today but will do ASAP, in meantime seeing what Travis says here https://travis-ci.org/gawkermedia/scribe-1/jobs/47131815

bregenspan:
So this was completely the fault of me relying on undocumented behavior that is apparently only present in the latest versions of Chrome (ability to set custom properties to an event), and not something exotic like Webdriver focus issues.  I can get things working by adding one solution that works for Chrome, one for FF, but have pretty mixed feelings about adding something that hacky.

An alternate idea is to introduce a little bit of ugliness in the paste handling by having it look for `event.clipboardData || event.detail.clipboardData` (the latter can safely be set in a custom event), but it's definitely a choice between two evils here, I'll keep thinking about it.

awentzonline:
Thanks for adding some tests to this, @bregenspan! Yeah it looks like this is the root of the chrome (and probably safari) problems: https://github.com/WebKit/webkit/blob/master/Source/WebCore/dom/Event.idl#L77-L79

bregenspan:
@awentzonline looking at the source is cheating!

So IMO the correct, least-hacky way of handling this is to go back to keyboard events like: https://github.com/gawkermedia/scribe-test-harness/compare/add-paste-helper-keyboard-version-3?expand=1 and just conditionally not execute paste-related tests on the OSX * Chrome combination until the Chromedriver bug with command keys is fixed.  That way there's still full coverage in Travis and local test-running on OSX still behaves at least in FF.

If that sounds good to everyone I'll try to fix the remaining glitches in that approach (e.g. https://travis-ci.org/gawkermedia/scribe-1/jobs/47267311) as soon as I can (but anyone should feel free to jump in if you have time and want to before then).

Sorry again for the fail to test outside of dev version of Chrome!

rrees:
@awentzonline @bregenspan @hmgibson23 and I we're discussing this the other day and the way we think is easiest to go forward is that we should bind the native cut and paste events to scribe internal events and have the paste functionality tests be based on the internal event.

We'll lack some coverage in the native event binding to the internal events but everything else would be reliably testable.

Let us know if you think this will work.

bregenspan:
@rrees I like that, feels like a clear way way! Means some paste event object might have to be mocked or there will need to be separate events for HTML vs plaintext but avoids this whole Selenium nightmare.

PR #320, opened 2015-01-15
Revert "Add support for paste formatters"

hmgibson23:
Reverts guardian/scribe#319


hmgibson23:
Reverting this because it's failing a test on Chrome. We had a brief look at the issue and it seems that Chrome can't insert the data in the custom event. We may look into this more in the future but it seems very close. And of course, it works fine in Firefox!

PR #321, opened 2015-01-15
Plugins: allow command patches to overridden

rrees:
This is a baby step towards allowing plugins to be configurable during scribe definition.

Primarily this aims to allow people to work around #250 in the simplest way. In this version the invoker can override the list of command patches they want and therefore not load things like the bold command patch.


hmgibson23:
Looks good :+1:

PR #323, opened 2015-01-21
Patched event emitter so multiple event assignment is imposible

cutandpastey:
Currently there is no check to see if an function has already been assigned to a particular event within `src/event-emitter.js`.  This causes problems within the noting plugin. Specifically when media is added to Composer as more scribe instances are created and thus event handlers are assigned multiple times, see: https://github.com/guardian/scribe-plugin-noting/issues/75.

This PR ensures event handlers can only be assigned to a particular event once.

### DIFF
--- a/src/event-emitter.js
+++ b/src/event-emitter.js
@@ -1,4 +1,4 @@
-define(['lodash-amd/modern/arrays/pull'], function (pull) {
+define(['lodash-amd/modern/arrays/pull', 'immutable/dist/immutable'], function (pull, Immutable) {
 
   'use strict';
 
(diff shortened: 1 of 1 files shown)


src/event-emitter.js
```diff
@@ -12,6 +12,12 @@ define(['lodash-amd/modern/arrays/pull'], function (pull) {
   EventEmitter.prototype.on = function (eventName, fn) {
     var listeners = this._listeners[eventName] || [];
 
+    listeners.forEach(function(listener){
+      if(listener === fn){
+        return;
+      }
+    });
```
theefer: I don't think this does what you expect. You're just returning from the function you passed to `forEach`, not the parent function `forEach` is called from. [`Array.prototype.some`](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Array/some) may be more useful here.

rrees:
What you've done is fine (barring @theefer's comment) but I think we can make this double-plus good by using ImmutableJS's Set collection which is exactly what we want without the explicit checking. We should also take the opportunity to switch to an Immutable map as the state mutation is clearly defined here.

theefer:
A Set makes complete sense here (ES6 has them ;-)). Not entirely sure what the benefit of immutability would be here, given it's mutated in place all the time?

rrees:
@theefer Unless I'm massively misunderstanding the code the listeners should be invoked frequently but not mutated except via the register/unregister functions in the EventEmitter.

theefer:
Yes that's right.

theefer:
Caught up with @rrees. Conclusion is: we agree that a Set is the simplest data structure to solve this, and ImmutableJS provides Sets, so it makes sense to use them. They are immutable, and we don't make heavy use of that here, but they're still sets, which is what we want.

PR #328, opened 2015-01-29
Event emitter: make callbacks unique

rrees:
I've applied the Set idea I mentioned in #323 to the code here.

I wanted to do a bit more but I hit an issue with unit tests that I'll share on the mailing list.

### DIFF
--- a/src/event-emitter.js
+++ b/src/event-emitter.js
@@ -11,24 +11,22 @@ define(['lodash-amd/modern/arrays/pull',
   }
 
   EventEmitter.prototype.on = function (eventName, fn) {
-    var listeners = this._listeners[eventName] || [];
+    var listeners = this._listeners[eventName] || Immutable.Set();
 
-    listeners.push(fn);
-
-    this._listeners[eventName] = listeners;
+    this._listeners[eventName] = listeners.add(fn);
   };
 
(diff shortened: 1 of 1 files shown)


cutandpastey:
I did this here https://github.com/guardian/scribe/pull/323/files after the comments, thought you would get a notification :(

rrees:
@cutandpastey Ah, I may have got the notifications but I may have missed it. At least the code is reassuring similar. I'll drop this PR.

rrees:
@cutandpastey Actually having a look at the change I'm not sure it works because you don't rewrite the var references, however it will probably work with the test suite because we only add one callback to the event emitter

rrees:
@cutandpastey Actually ignore me they both look the same

cutandpastey:
You `clear()` the set.  Should I add this in or are we all good?

hmgibson23:
The build also failed! Maybe run again?

rrees:
@cutandpastey the idea with the clear is that the structure stays the same but empty

cutandpastey:
Yeah, timeout issue :( have re-run.

cutandpastey:
@rrees Oh, I thought this was for GC.

Issue #308, opened 2014-12-23
Behavior when inserting contenteditable="false" elements


christopherliu:
I'm using contenteditable="false" elements in my editor pane to represent atomic elements (a chart, for example, might have a DOM but we'd want to delete it all at once). While I was looking at the code, I noticed that certain areas use the existence of the contenteditable attribute as a proxy for the Scribe container.

https://github.com/guardian/scribe/blob/master/src/api/selection.js:23
https://github.com/guardian/scribe/blob/master/src/api/node.js:15

theefer:
> just select the scribe element itself

:+1:

rrees:
@christopherliu I believe this should be fixed in 1.2.4

PR #331, opened 2015-02-02
API: use the Scribe root element to stop DOM traversal

rrees:
This picks up #308 and tries to use the element that the Scribe instance is bound to determine whether the searched element is at the root or not.

I'm not sure how much I trust the test suite on this but things are not broken &trade;


hmgibson23:
LGTM :+1:

rrees:
@hmgibson23 I'll merge tomorrow morning, I'm not sure I've got time to do the full release process now

PR #332, opened 2015-02-08
Fixing a bug caused when you automatically trigger focus

gja:
to scribe controlled DOM

How I'm triggering this:
```js
      var range = document.createRange();
      range.selectNodeContents(element);
      range.collapse(false);
      var selection = window.getSelection();
      selection.removeAllRanges();
      console.log(range);
      selection.addRange(range);
```

### DIFF
--- a/src/plugins/core/events.js
+++ b/src/plugins/core/events.js
@@ -41,7 +41,7 @@ define([
           var isFirefoxBug = scribe.allowsBlockElements() &&
                   selection.range.startContainer === scribe.el;
 
-          if (isFirefoxBug) {
+          if (isFirefoxBug && scribe.el.firstChild) {
             var focusElement = getFirstDeepestChild(scribe.el.firstChild);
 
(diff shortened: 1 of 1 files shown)


hmgibson23:
This seems sane. Could you possibly make a jsbin with the bug reproduced?

gja:
Is there an existing jsbin I can duplicate or something?

rrees:
@gja This was one I found from an earlier issue:

http://jsfiddle.net/8g6ezbsb/3/

hmgibson23:
@gja  - There is also this one from the brower inconsistencies file http://jsbin.com/iwEWUXo/2/edit?js,console,output

gja:
Thanks, I've been a bit busy, but i'll create this later today

hmgibson23:
Great. If you add to the browser inconsistencies file then we can keep track of it...

rrees:
Closing as inactive.

PR #333, opened 2015-02-09
Add additional function arguments to document.createTreeWalker calls

deains:
This is a small fix for IE compatibility. IE doesn't consider the arguments to createTreeWalker optional, so it bombs out if any are missing. This is causing observeDomChanges to break in IE11, and may well be breaking other stuff too.

Adding in the arguments with their default values should not cause any compatibility issues with other browsers.

See here for reference: https://developer.mozilla.org/en-US/docs/Web/API/Document.createTreeWalker


hmgibson23:
Looks fins to me :+1:

PR #337, opened 2015-02-12
Added event namespacing

cutandpastey:
Adding namespacing to events will allow us to cleanup some of the state management in noting.


.gitignore
```diff
@@ -4,3 +4,4 @@
 /vendor/
 /examples/build.js
 .tern-port
+.npm-degub.log
```
theefer: should it be "debug"?

src/event-emitter.js
```diff
@@ -26,11 +26,18 @@ define(['lodash-amd/modern/arrays/pull',
...
 
   EventEmitter.prototype.trigger = function (eventName, args) {
-    var listeners = this._listeners[eventName] || Immutable.Set();
 
-    listeners.forEach(function (listener) {
-      listener.apply(null, args);
-    });
+    //fire events like my:custom:event -> my:custom -> my
+    var events = eventName.split(':');
+    while(events.length){
```
theefer: Please avoid using integers as booleans for clarity, [as per styleguide](https://github.com/guardian/flexible-content/wiki/JavaScript-Style-Guide#syntax).

theefer:
Not entirely sure I understand the use case, but the code looks right! :+1:

src/event-emitter.js
```diff
@@ -26,11 +26,18 @@ define(['lodash-amd/modern/arrays/pull',
...
+      var currentEvent = events.join(':');
+      var listeners = this._listeners[currentEvent] || Immutable.Set();
```
rrees: I'm not sure this is quite right, I think the scope of this statement (due to the initialisation) is outside the while loop? We don't want the listeners to vary during the execution right?

rrees:
@cutandpastey Excuse my ignorance but why do you want the sub-events fired?

If I pass in 'scribe:content-changed' why would I want to fire an event called 'scribe'? 

In my mind this was more that the Scribe EventEmitter would add a namespace to the events being fired if one wasn't provided. So if I sent 'hello' to the Scribe EventEmitter then it would emerge as 'scribe:hello'

rrees:
Good work on getting unit-testing working, it's going to make things much easier in future :clap:

.gitignore
cutandpastey: Doh!!

cutandpastey:
So in noting I want to fire something like `note:toggle-all:action`. This way all the noting controllers can collapse/toggle all their respective notes and I can have ONE handler on `note:toggle-all` which inverts the note collapse state. This will contain the actions within the controllers and contain state management to the global.

rrees:
@cutandpastey Okay I understand so this is actually like a fountain event rather than namespacing?

hmgibson23:
:+1:

rrees:
:+1: with some wiki documentation and a green build!

cutandpastey:
Is it me or is it impossible to re-run a travis build without pushing a commit?

rrees:
@cutandpastey You should see a refresh button on your "Job" status page you need to have logged in via your Github account to get the permissions though

cutandpastey:
@rrees let me know if this is all good: https://github.com/guardian/scribe/wiki/Components

Now its time to mash that Travis button

rrees:
@cutandpastey Looks good :shipit:

Issue #329, opened 2015-01-29
Fails to retain focus due to invalid range comparisons when in ShadowDOM


bigdave:
While trying to get Scribe to work within a Polymer-based application, we ran into a problem. The proof-of-concept polymer element for Scribe keeps the editor in the light DOM, which is fine for that example, but when using the editor within a Polymer component's Shadow DOM, the editor fails to keep focus, and throws exceptions.  It appears that it is attempting to compare ranges, assuming that the editor is present in `window.document`, when in fact is resides in the Shadow DOM document of the Polymer element.

This results in the error `Uncaught WrongDocumentError: Failed to execute 'compareBoundaryPoints' on 'Range': The source range is in a different document than this range. selection.js:52`.

I have setup a [proof of concept](http://bigdave.github.io/scribe-editor-element/) for this error, which must be viewed in a browser which actually has a Web Components Shadow DOM implementation such as Chrome, Opera, or Vivaldi. The behavior will not appear in FireFox (unless you're using a fancy nightly with the toggle turned on).

bigdave:
If you're looking at scribe.js as opposed to selection.js, the error is occurring on line 3844

hmgibson23:
Not sure but this could be mitigated by using ```ownerDocument``` to get the selection. However, when doing this we ran into some issues in FF.

shaunnetherby:
This issue is affecting me as well.  `ownerDocument` seems to be the same as `window.document`.  In Chrome, if you grab the shadowRoot of the target element (`el`) and call getSelection on that root, the selection object has the correct nodes (which means it is in the correct document fragment).  However when `getRangeAt` is called the range object has a common ancestor in the `window.document` instead of the shadowRoot.  This results in the compareBoundaryPoints method throwing an error because `this.range` is in the `window.document` and `scribeNodeRange` is in the shadowRoot document.

      var selectionStartWithinScribeElementStart = this.range.compareBoundaryPoints(Range.START_TO_START, scribeNodeRange) >= 0;

xRapidDavex:
I'm having this same problem. Can we get an ETA on a fix?

hmgibson23:
Is an issue with Polymer?

miztroh:
Perhaps it's related to this Chrome bug: https://code.google.com/p/chromium/issues/detail?id=380690

hmgibson23:
It might be. If it's an issue with Polymer only, it's highly unlikely we'll provide a fix for this.

[...]

PR #338, opened 2015-02-12
Fix/329 chrome shadow dom errors

shaunnetherby:
This is to fix issue #329. Scribe does not work in shadow DOM because it grabs the selection from the document rather then the parent document fragment (shadow root in this case).  I updated Selection to grab the appropriate document fragment to use to get the selection.

There is also 2x chrome bugs encountered when scribing in shadow root.  The first causes the selection.getRangeAt() method to return an incorrect range.  I checked Chrome Canary and found that the getRangeAt() issue is fixed in that version but I applied the workaround so scribe can be used for now (https://code.google.com/p/chromium/issues/detail?id=380690).

The second bug is not fixed in Canary and it causes selection.isCollapsed to be incorrect.  I found a workaround by calling range.collapsed (https://code.google.com/p/chromium/issues/detail?id=447523)

I tried to create some tests for these cases but was having trouble.  I think the scribe-test-harness will have to be modified to support testing scribe in a shadow root, and I'm not just not sure how to appropriately do that.  All of the current tests did pass in FF/Chrome.

### DIFF
--- a/src/api/selection.js
+++ b/src/api/selection.js
@@ -14,12 +14,12 @@ function (elementHelper) {
 
       // find the parent document or document fragment
       var currentElement = scribe.el.parentNode;
-      while(currentElement && currentElement.nodeType !== 11 && currentElement.nodeType !== 9) { // 11 = DOCUMENT_FRAGMENT_NODE, 9 = DOCUMENT
+      while(currentElement && currentElement.nodeType !== Node.DOCUMENT_FRAGMENT_NODE && currentElement.nodeType !== Node.DOCUMENT_NODE) {
         currentElement = currentElement.parentNode;
       }
 
       // if we found a document fragment and it has a getSelection method, set it to the root doc
-      if (currentElement.nodeType === 11 && currentElement.getSelection) {
+      if (currentElement.nodeType === Node.DOCUMENT_FRAGMENT_NODE && currentElement.getSelection) {
         rootDoc = currentElement;
       }
 
--- a/src/api/selection.js
+++ b/src/api/selection.js
@@ -10,10 +10,26 @@ function (elementHelper) {
      * Wrapper for object holding currently selected text.
      */
     function Selection() {
-      this.selection = window.getSelection();
+      var rootDoc = document;
 
+      // find the parent document or document fragment
+      var currentElement = scribe.el.parentNode;
+      while(currentElement && currentElement.nodeType !== 11 && currentElement.nodeType !== 9) { // 11 = DOCUMENT_FRAGMENT_NODE, 9 = DOCUMENT
+        currentElement = currentElement.parentNode;
+      }
+
+      // if we found a document fragment and it has a getSelection method, set it to the root doc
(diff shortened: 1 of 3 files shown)


rrees:
@shaunnetherby As long as the tests pass I'd be happy to merge this, could you add the issues you're working around to BROWSERINCONSISTENCIES.md as well so we remember to review them later.

rrees:
Second opinion @cutandpastey ?

cutandpastey:
Looks good. Tests pass in both Chrome and FF here. I don't have a lot of experience with shadow DOM so maybe @theefer should cast a glance. :+1: from me tho.

src/api/selection.js
```diff
@@ -10,10 +10,26 @@ function (elementHelper) {
      * Wrapper for object holding currently selected text.
      */
     function Selection() {
-      this.selection = window.getSelection();
+      var rootDoc = document;
 
+      // find the parent document or document fragment
+      var currentElement = scribe.el.parentNode;
+      while(currentElement && currentElement.nodeType !== 11 && currentElement.nodeType !== 9) { // 11 = DOCUMENT_FRAGMENT_NODE, 9 = DOCUMENT
```
theefer: Please use constants on `Node` rather than ids that then need to be explicited by comments.

For example `Node.DOCUMENT_FRAGMENT_NODE` is `11`, etc.

theefer:
The code seems sane but I don't really have enough context from working on Scribe recently to know what this is really doing, so would be good to confirm with someone QA'ing it and also even some tests if possible?

@shaunnetherby Are you building a Scribe Web Component? Is it available to see anywhere? Would be really interested to see how well it works and the kind of HTML interface it exposes!

cutandpastey:
I think there are a few people doing this? https://github.com/guardian/scribe/issues/329

shaunnetherby:
@theefer Good catch on using the Node constants, I updated the codes accordingly.  I think tests would be great but I'm not sure how to introduce shadow dom testing into the current test framework.  I played around with it a bit but I think I would need to do some serious tweaking in the scribe-test-harness helpers.js file, and I'm a little hesitant to play with that file since it affects so many of the tests.  Ideally there would be a way that this could be tweaked such that all of the tests could be run against a light dom scribe.el and a shadow dom scribe.el to root out any other shadow inconsistencies.

@rrees I updated the BROWSERINCONSISTENCIES.md in a way that I think makes sense, so take a look at that.

I'll try to post a link here by this afternoon so you can check out how I'm using scribe (need to get some stuff in order first).

shaunnetherby:
I added a missing null check for currentElement (was causing issues in some of my tests).

@theefer Here is a link to the component I am working on, should be able to start making more progress now that scribe is behaving :).
http://infusionsoft.github.io/inf-rich-text-editor/bower_components/inf-rich-text-editor/demo.html

rrees:
@shaunnetherby Awesome work I (or someone else) will merge and release as soon as we can. Good luck with the Web Component!

rrees:
@shaunnetherby Scribe 1.2.7

PR #340, opened 2015-02-17
Superficial (non-functionality affecting) changes

christopherliu:
When creating my own Commands, I had to search through the codebase to determine that nodeName was not being used externally; this small visual change helps to make it clear that implementors of Command need not implement nodeName.


rrees:
Looks good to me, the fn parameter in the formatter should probably just be called formatter but I'll clean that up later.

Thanks! :+1:

PR #341, opened 2015-02-18
Formatters: rename parameter

rrees:
This is pretty trivial the parameter name *fn* while true was needlessly general


hmgibson23:
:+1:

PR #343, opened 2015-02-20
Selection API range property is incorrect for "backwards" selection

deains:
This is a bug which looks to have been introduced by [edd638f], where `getRangeAt(0)` was replaced with a `createRange` routine.

The issue is that if you select some text backwards, i.e. anchorNode is after focusNode, the range will be empty, because although Selection is "backwards"-compatible, Range is not.

Here's a JSbin which demonstrates the issue: http://jsbin.com/wujuropeye/1/edit?html,js,output

### DIFF
--- a/src/api/selection.js
+++ b/src/api/selection.js
@@ -28,8 +28,19 @@ function (elementHelper) {
         // create the range to avoid chrome bug from getRangeAt / window.getSelection()
         // https://code.google.com/p/chromium/issues/detail?id=380690
         this.range = document.createRange();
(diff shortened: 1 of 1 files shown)


hmgibson23:
Seems fine to me @rrees second pair of eyes?

rrees:
Looks good, I think we need some tests around this though. Happy to accept the pull request as is though as it has the JSbin example.

rrees:
Thanks, released in v1.2.10

PR #344, opened 2015-02-23
Added "shouldUndo" to scribe options for optional undo management

cutandpastey:
As mentioned here: https://github.com/guardian/scribe-plugin-advanced-undo/pull/1 I've tried to make scribe undo functionality optional by configuration. Not 100% sure of all the impacts of this so would be good to get eyes @rrees @hmgibson23 @robinedman @jamespamplin @theefer  

NB: all the tests run locally.

### DIFF
--- a/src/plugins/core/events.js
+++ b/src/plugins/core/events.js
@@ -92,9 +92,9 @@ define([
           // `scribe.setContent`), we do not want to push to the history. (This
           // happens on the first `focus` event).
           if (isEditorActive) {
-            // Discard the last history item, as we're going to be adding
-            // a new clean history item next.
             if (scribe.options.shouldUndo) {
+              // Discard the last history item, as we're going to be adding
+              // a new clean history item next.
               scribe.undoManager.undo();
             }
 
--- a/src/plugins/core/events.js
+++ b/src/plugins/core/events.js
@@ -92,7 +92,7 @@ define([
           // `scribe.setContent`), we do not want to push to the history. (This
           // happens on the first `focus` event).
           if (isEditorActive) {
-            if (scribe.options.shouldUndo) {
+            if (scribe.undoManager) {
               // Discard the last history item, as we're going to be adding
               // a new clean history item next.
               scribe.undoManager.undo();
--- a/src/plugins/core/events.js
+++ b/src/plugins/core/events.js
@@ -94,7 +94,9 @@ define([
           if (isEditorActive) {
             // Discard the last history item, as we're going to be adding
             // a new clean history item next.
-            scribe.undoManager.undo();
+            if (scribe.options.shouldUndo) {
+              scribe.undoManager.undo();
+            }
 
             // Pass content through formatters, place caret back
             scribe.transactionManager.run(runFormatters);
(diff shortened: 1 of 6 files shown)


theefer:
Sounds like we want to identify the behaviour that is achieved by interacting with the undo manager in these patches (writing test for these cases if possible) before we make it optional?

cutandpastey:
I don't feel that is necessary. The only two places it's not totally encapsulated is in `src/plugins/core/events.js` & `src/plugins/core/patches/events.js` where the call to undo is there to clear the last undo item before formatters run. If undo is optional then this clearly isn't required.

hmgibson23:
I'm fine with this :+1:

theefer:
OK so no need to cleanup the UndoManager's history because the UndoManager is completely bypassed? Fair enough.

Seems like we'll want a cleaner way to factor out the current undo manager and replace it with another one, in the future.

theefer:
:+1:

src/plugins/core/events.js
```diff
@@ -94,7 +94,9 @@ define([
           if (isEditorActive) {
             // Discard the last history item, as we're going to be adding
```
rrees: Comment should now be inside the if-block

src/scribe.js
```diff
@@ -47,6 +47,7 @@ define([
     this.options = defaults(options || {}, {
       allowBlockElements: true,
       debug: false,
+      shouldUndo: true,
```
rrees: It's a small thing but could this be {undo: {enabled: <true|false>}}?

src/scribe.js
```diff
@@ -71,8 +72,10 @@ define([
     var TransactionManager = buildTransactionManager(this);
     this.transactionManager = new TransactionManager();
 
-    var UndoManager = buildUndoManager(this);
-    this.undoManager = new UndoManager();
+    if (this.options.shouldUndo) {
+      var UndoManager = buildUndoManager(this);
+      this.undoManager = new UndoManager();
```
rrees: I feel undoManager should be `undefined` and then take on a value if the options say it should do. Does that make sense?

cutandpastey: This will be the case anyway. For example `window.myAwesomeVar === undefined`. Unless I'm missing your point?

src/scribe.js
cutandpastey: Im not sure thats necessary as there are no more configuration options to pass.  If there were then I'd be up for this.

src/scribe.js
theefer: Oh I didn't notice that. Could the presence of `scribe.undoManager` be used in conditions instead of `scribe.options.shouldUndo` in that case? That would reduce the API surface the code has to know about (test for the presence of the thing you're about to conditionally use, rather than for a flag that you know implicitly determines whether that thing you're about to conditionally use has been set).

cutandpastey: In the cases where `scribe.undoManager` is called directly this makes sense, but thats not the only pieces of functionality that are changed. The assignment of event listeners are determined by this option and it makes no practical sense to check against the undoManager in this case. In order to be consistent therefore, we should always check against the option.

theefer: True, :+1:

cutandpastey:
@rrees made all requested amends.

rrees:
:+1:

Issue #335, opened 2015-02-11
"Accelerating" undo


rrees:
Rather than undoing character by character users would like to switch from character to word-based undo if they are repeatedly hitting the undo key combo.

robinedman:
It's worth noting that we don't technically undo character by character. Rather we undo the last "transaction" (managed by Scribe's transaction manager), which could be more e.g. if you select and then erase a paragraph then that would be one transaction.

Example:
![scribe-undo](https://cloud.githubusercontent.com/assets/1936213/6153035/81a423dc-b219-11e4-8edd-3a559fc98fe6.gif)

I wonder how other text editors solve this. It's clearly useful but it might make Scribe more complex.

theefer:
@OliverJAsh tried to understand how OS/browsers do it but there was no obvious pattern. One way would be to debounce the history recording. There may or may not be subtle implications, given that some of the hack applied to normalise content editable across browsers sometimes have to fiddle with the history stack IIRC.

But featurewise, it would be nice to be closer to native undo, whatever the hell that actually is.

sihil:
Suggest someone looks at the behaviour of InCopy as that seems the logical thing to copy.

theefer:
![horse-cart-car](https://cloud.githubusercontent.com/assets/36964/6158151/be970d92-b23f-11e4-8ff3-1e77ad0ec8f5.jpg)
Beware the obscure consequences of legacy on modernity :-)

blishen:
Those who fail to learn from history are doomed to repeat it

On 11 February 2015 at 22:46, Sébastien Cevey <notifications@github.com>
wrote:

> [image: horse-cart-car]
> <https://cloud.githubusercontent.com/assets/36964/6158151/be970d92-b23f-11e4-8ff3-1e77ad0ec8f5.jpg>
> Beware the obscure consequences of legacy on modernity :-)
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/guardian/scribe/issues/335#issuecomment-73981753>.
>



-- 
PGP Public Key <http://bit.ly/Rxbzi0>

-- 

------------------------------
Visit theguardian.com. On your mobile and tablet, download the Guardian 
iPhone and Android apps theguardian.com/guardianapp and our tablet editions 
theguardian.com/editions.  Save up to 57% by subscribing to the Guardian 
and Observer - choose the papers you want and get full digital access. 
 Visit subscribe.theguardian.com

This e-mail and all attachments are confidential and may also be 
[...]

Issue #334, opened 2015-02-11
Shared undo manager


aaalsaleh:
It would be great if the undo manager could be shared across multiple instances. One way to do that is to make the undo manager as a plugin that can be instantiated by the user once and used by multiple scribe instances (shared), or an instance for each scribe instance (exclusive). To make this work, each undo transaction should keep a reference to the scribe instance it belongs to (source of transaction).

This can also allow the undo manager to be extended and used by other non-scribe components on the same page. Of course, these components need to define and push their own transactions according to the undo manager API. This leads to a more streamlined undo manager across the whole page or web application.

rrees:
On the face of it that changes looks like it might be very complicated, do you have an example of why something external would want to manipulate the undo log?

I think it might be better if we allowed you to specify a replacement for the default undo manager rather than trying to implement this ourselves on your behalf.

aaalsaleh:
That is basically the idea, making the undo manager an external plugin. It is not necessary that external components could manipulate the undo stack. As you stated, this can be done by those who needs it once it is an exchangeable plugin. The important thing is at least allowing the possibility for sharing the undo stack between one or more scribe instances.

One example for scribe specifically: two or more instances of scribe on the same page (e.g. summary and full article) that can be edited simultaneously. You can make edits here and there interchangeably. It would be useful for some applications if the undo stack was shared between them as if it was a single document.

Another example for scribe with other components: one or more scribe instances along side other input components in a web application. The user could implement their own global undo plugin that can handle both scribe and other components and use it as scribe's undo manager. This will allow the user to implement a global undo manager and use it with scribe.

rrees:
Okay, if its urgent I'd suggest submitting a PR otherwise at some point I'll look at making the transaction manager pluggable

aaalsaleh:
Here you go

rrees:
@aaalsaleh :clap: :clap: :clap: :cake:

aaalsaleh:
@rrees I think we can finally consider this implemented :wink:

PR #346, opened 2015-03-04
UndoManager Improvements

aaalsaleh:
This pull request implements #334, and provides a form of improvement for the undo manager behavior as suggested in #335:

1. Replaces the current undo manager with a new undo manager. The new undo manager follows the W3C UndoManager specifications, and is more flexible than the current one. See: https://github.com/aaalsaleh/undo-manager
2. Updates the way transactions are pushed into the new undo manager.
3. Adds an `interval` option for the undo manager. Now, consecutive transactions within `interval` will be merged as a single transaction (#335).
4. ~~Adds a new `setUndoManager` API for replacing the internal undo manager with external undo manager (#334).~~
5. Several other tiny updates over the undo/redo core commands, and core/patch events. These are needed for the new undo manager.

Please note that point 2 and 3 can be further improved and optimized specifically for Scribe. For example, ~~transactions content can be replaced instead of merged to save space~~ (done!). Also, merging condition can be improved to take other factors into account.

### DIFF
--- a/src/scribe.js
+++ b/src/scribe.js
@@ -47,7 +47,7 @@ define([
     this.options = defaults(options || {}, {
       allowBlockElements: true,
       debug: false,
-      undo: { enabled: true, limit: 100, interval: 1000 },
+      undo: { manager: null, enabled: true, limit: 100, interval: 1000 },
       defaultCommandPatches: [
         'bold',
         'indent',
@@ -75,11 +75,18 @@ define([
     //added for explicit checking later eg if (scribe.undoManager) { ... }
     this.undoManager = false;
     if (this.options.undo.enabled) {
-      this.undoManager = new UndoManager(this.options.undo.limit, this.el);
+      if (this.options.undo.manager) {
+        this.undoManager = this.options.undo.manager;
+      }
+      else {
+        this.undoManager = new UndoManager(this.options.undo.limit, this.el);
+      }
       this._merge = false;
       this._forceMerge = false;
       this._mergeTimer = 0;
     }
+
+    this.setHTML(this.getHTML());
 
     this.el.setAttribute('contenteditable', true);
 
@@ -168,7 +175,6 @@ define([
     }
 
     this._previousContent = this.getHTML();
-    console.log(this._previousContent);
   };
 
   Scribe.prototype.getHTML = function () {
--- a/src/scribe.js
+++ b/src/scribe.js
@@ -50,7 +50,7 @@ define([
     this.options = defaults(options || {}, {
       allowBlockElements: true,
       debug: false,
-      undo: { enabled: true, interval: 1000 },
+      undo: { enabled: true, interval: 0, limit: 100 },
       defaultCommandPatches: [
         'bold',
         'indent',
@@ -78,7 +78,7 @@ define([
     //added for explicit checking later eg if (scribe.undoManager) { ... }
     this.undoManager = false;
     if (this.options.undo.enabled) {
-      this.undoManager = new UndoManager(100, this.el);
+      this.undoManager = new UndoManager(this.options.undo.limit, this.el);
     }
 
     this.el.setAttribute('contenteditable', true);
@@ -195,7 +195,7 @@ define([
        */
 
       // We only want to push the history if the content actually changed.
-      if (! previousUndoItem || (previousUndoItem && this.getHTML() !== previousContentNoMarker)) {
+      if (this.undoManager.length == 0 || this.getHTML() !== previousContentNoMarker) {
         var selection = new this.api.Selection();
 
         selection.placeMarkers();
--- a/src/scribe.js
+++ b/src/scribe.js
@@ -84,6 +84,7 @@ define([
       this._merge = false;
       this._forceMerge = false;
       this._mergeTimer = 0;
+      this._lastItem = {content: ''};
     }
 
     this.setHTML(this.getHTML());
@@ -166,6 +167,8 @@ define([
   };
 
   Scribe.prototype.setHTML = function (html, skipFormatters) {
+    this._lastItem.content = html;
+
     if (skipFormatters) {
       this._skipFormatters = true;
     }
@@ -173,8 +176,6 @@ define([
     if (this.el.innerHTML !== html) {
       this.el.innerHTML = html;
     }
-
-    this._previousContent = this.getHTML();
   };
 
   Scribe.prototype.getHTML = function () {
@@ -200,38 +201,38 @@ define([
 
     if (scribe.options.undo.enabled) {
       // Get scribe previous content, and strip markers.
-      var previousContent = scribe._previousContent;
-      var previousContentNoMarkers = previousContent && previousContent
+      var lastContentNoMarkers = scribe._lastItem.content
         .replace(/<em class="scribe-marker">/g, '').replace(/<\/em>/g, '');
 
       // We only want to push the history if the content actually changed.
-      if (scribe.getHTML() !== previousContentNoMarkers) {
+      if (scribe.getHTML() !== lastContentNoMarkers) {
         var selection = new scribe.api.Selection();
 
         selection.placeMarkers();
         var content = scribe.getHTML();
         selection.removeMarkers();
 
-        // Checking if there is a need to merge, there is an item to merge with,
-        // and the last item is a scribe transaction of the same source.
+        // Checking if there is a need to merge, and that the previous history item
+        // is the last history item of the same scribe instance.
         // It is possible the last transaction is not for the same instance, or
         // even not a scribe transaction (e.g. when using a shared undo manager).
-
-        var lastItem = scribe.undoManager.item(scribe.undoManager.position);
-        if ((scribe._merge || scribe._forceMerge) && lastItem && scribe == lastItem[0].scribe) {
-          // Merge manually with the last item to save more memory space.
-          lastItem[0].content = content;
+        var previousItem = scribe.undoManager.item(scribe.undoManager.position);
+        if ((scribe._merge || scribe._forceMerge) && previousItem && scribe._lastItem == previousItem[0]) {
+          // If so, merge manually with the last item to save more memory space.
+          scribe._lastItem.content = content;
         }
         else {
-          // Otherwise, register a new transaction.
-          scribe.undoManager.transact({
-            previousContent: previousContent,
+          // Otherwise, create a new history item, and register it as a new transaction
+          scribe._lastItem = {
+            previousItem: scribe._lastItem,
             content: content,
             scribe: scribe,
             execute: function () { },
-            undo: function () { this.scribe.restoreFromHistory(this.previousContent); },
-            redo: function () { this.scribe.restoreFromHistory(this.content); }
-          }, false);
+            undo: function () { this.scribe.restoreFromHistory(this.previousItem); },
+            redo: function () { this.scribe.restoreFromHistory(this); }
+          };
+
+          scribe.undoManager.transact(scribe._lastItem, false);
         }
 
         // Merge next transaction if it happens before the interval option, otherwise don't merge.
@@ -251,7 +252,9 @@ define([
   };
 
   Scribe.prototype.restoreFromHistory = function (historyItem) {
-    this.setHTML(historyItem, true);
+    this._lastItem = historyItem;
+
+    this.setHTML(historyItem.content, true);
 
     // Restore the selection
     var selection = new this.api.Selection();
--- a/src/scribe.js
+++ b/src/scribe.js
@@ -29,7 +29,7 @@ define([
   setRootPElement,
   Api,
   buildTransactionManager,
-  buildUndoManager,
+  UndoManager,
   EventEmitter,
   elementHelpers,
   nodeHelpers,
@@ -37,6 +37,9 @@ define([
 ) {
 
   'use strict';
+
+  var intervalTimer = 0;
+  var merge = false;
 
   function Scribe(el, options) {
     EventEmitter.call(this);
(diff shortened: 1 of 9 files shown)


aaalsaleh:
The build failed in e65b169 due to `Uncaught UnknownError: Sauce Labs Authentication Error`. Otherwise, the commit is correct and exactly like 00e08f2, which didn't fail. There is no way to force rebuild without another commit or closing and reopening this PR...

src/scribe.js
```diff
@@ -179,10 +181,19 @@ define([
     return this.el.textContent;
   };
 
+  /**
+   * Replace the current undo manager. Can be used by undo manager plugins or directly by the user.
+   * @param {Object} undoManager The undo manager instance to be used by the current editor.
+   */
+  Scribe.prototype.setUndoManager = function (undoManager) {
+	  this.undoManager = undoManager;
```
hmgibson23: We use two spaces for alignment in scribe. You should be able use to editorconfig to sort this out for you...

src/undo-manager.js
```diff
@@ -1,43 +1,73 @@
...
-        console.log('UndoManager.push: %s', item);
-      }
-      this.stack.length = ++this.position;
-      this.stack.push(item);
+			stack.splice(0, this.position);
+			if (merge && this.length)
+				stack[0].push(transaction);
+			else
+				stack.unshift([transaction]);
+			this.position = 0;
```
hmgibson23: Indentation is quite strange again from line 15 - 22.

cutandpastey:
Hey, Thanks for the PR! In general I like the approach there are, however, a few issues:

- All the undo-manager tests in `app/undo-manager.spec.js` fail for me here.
- There are several style issues throughout the code, mainly indentation and brackets around `if` statements.
- Personally I'd like to see the `UndoManager` methods moved onto the prototype rather than being defined within the constructor, but this is opinion. Not sure how @hmgibson23 @rrees and @robinedman feel about this?

If you can do a quick tidy and refactor the tests I'll get this merged ASAP!

src/undo-manager.js
```diff
@@ -1,43 +1,73 @@
...
 
-      while (this.stack.length > this.maxStackSize) {
-        this.stack.shift();
-        --this.position;
-      }
-    };
+			if (limit && stack.length > limit)
+				this.length = stack.length = limit;
+			else
+				this.length = stack.length;
```
hmgibson23: We alway use curly braces in for our if statements

src/undo-manager.js
```diff
@@ -1,43 +1,73 @@
...
 
-    UndoManager.prototype.undo = function () {
-      if (this.position > 0) {
-        return this.stack[--this.position];
-      }
-    };
+			if (fireEvent)
+				undoScopeHost.dispatchEvent(new CustomEvent('DOMTransaction', {detail: {transactions: stack[0].slice()}, bubbles: true, cancelable: false}));
```
hmgibson23: Indentation and curly braces

src/undo-manager.js
```diff
@@ -1,43 +1,73 @@
...
-    UndoManager.prototype.redo = function () {
-      if (this.position < (this.stack.length - 1)) {
-        return this.stack[++this.position];
-      }
-    };
+		this.undo = function () {
+			if (this.position < this.length) {
+				for (var i = stack[this.position].length - 1; i >= 0; i--)
+					stack[this.position][i].undo();
+				this.position++;
```
hmgibson23: Can you wrap this for with curly braces. Also the indentation is off

src/undo-manager.js
```diff
@@ -1,43 +1,73 @@
...
 
-    return UndoManager;
-  };
+				if (fireEvent)
+					undoScopeHost.dispatchEvent(new CustomEvent('undo', {detail: {transactions: stack[this.position - 1].slice()}, bubbles: true, cancelable: false}));
```
hmgibson23: Curly braces and indentation.

src/undo-manager.js
```diff
@@ -1,43 +1,73 @@
...
+			stack.length = this.length = this.position;
+		};
+
+		this.clearRedo = function () {
+			stack.splice(0, this.position);
+			this.position = 0;
+			this.length = stack.length;
+		};
+	}
+
```
hmgibson23: Indentation above.

hmgibson23:
This looks solid to me. It's definitely better than our current solution. I listed a few style gripes but we should probably add a style guide anyway. I'm also with @cutandpastey about moving the methods onto the prototype as it's more idiomatic.

src/scribe.js
```diff
@@ -179,10 +181,19 @@ define([
     return this.el.textContent;
   };
 
+  /**
+   * Replace the current undo manager. Can be used by undo manager plugins or directly by the user.
+   * @param {Object} undoManager The undo manager instance to be used by the current editor.
+   */
+  Scribe.prototype.setUndoManager = function (undoManager) {
```
rrees: I don't think it should be possible to change the UndoManager outside initialisation. Not a blocker to merging but not something we should support either.

src/scribe.js
```diff
@@ -192,14 +203,28 @@ define([
...
+      if (! previousUndoItem || (previousUndoItem && this.getHTML() !== previousContentNoMarker)) {
         var selection = new this.api.Selection();
 
         selection.placeMarkers();
         var html = this.getHTML();
         selection.removeMarkers();
 
-        this.undoManager.push(html);
+        // Register a new change transaction.
+        var scribe = this;
```
rrees: Not a blocker but having bound this to a name I think it would be better to use it on the next line as well

src/undo-manager.js
rrees: When we were discussing this internally we thought it was better to use a reversed stack (append at head) rather than a decrementing index loop.

rrees:
@aaalsaleh This is amazing! Thanks so much! :shipit:

src/undo-manager.js
```diff
@@ -1,43 +1,84 @@
...
 
-    UndoManager.prototype.redo = function () {
-      if (this.position < (this.stack.length - 1)) {
-        return this.stack[++this.position];
-      }
-    };
+		this.undo = function () {
+			if (this.position < this.length) {
+				for (var i = stack[this.position].length - 1; i >= 0; i--) {
+					stack[this.position][i].undo();
```
hmgibson23: So if you have a reversed append as @rrees suggested, you could just a forEach here.

aaalsaleh:
@hmgibson23 Thanks for pointing out these weird indentations. It was caused by my vim config.

@cutandpastey 
* It could be just due to the fact the undo manager was replaced with a different one. Could be also due to the behavior of the helpers used in the tests. Of course, it could be that I didn't handle all the cases properly while updating the code.
* I fixed all the coding style issues. Kindly point out if I missed any of them.
* The main reason I chose the constructor approach, is to have the stack and other options private. I don't think it is possible to have private properties with prototype approach. Nonetheless, I can move everything into prototype by making the private variables public, and just prepending them with `_` if needed.

@rrees I removed `setUndoManager` as noted. I originally intended to add the undo manager as a constructor option, but was not sure this will be accepted. I really didn't get your other remark about using the named `this` in the line below...

src/scribe.js
```diff
@@ -192,14 +195,28 @@ define([
...
+        this.undoManager.transact({
```
cutandpastey: As in `scribe.undoManager` rather than `this.undoManager`.

cutandpastey:
@aaalsaleh `undoManager.stack` was public before your amends so I don't think you need to worry about that being private and I can't see the value of hiding the check either.

aaalsaleh:
@cutandpastey Yes, I know. But it is not only hiding the check, it is about hiding the whole stack from direct modifications. They were moved to prototype earlier.

I am still unfamiliar with the details of how scribe works internally, and therefore, I would suggest extensively testing all the special cases. I removed some lines from core/events and patch/events, that were causing issues with the new undo manager. I am not quite sure what special cases they were handling, but without them the new undo manager works as expected. Please make sure you don't merge something adding such feature, only to break the main feature of the editor later on!

rrees:
@aaalsaleh I've copied your branch to my local machine and it looks like there is just one failing test now where the undo manager restores the uncleaned html (with style attributes) rather than the cleaned.

aaalsaleh:
@cutandpastey Turned out the tests in `test/undo-manager.spec.js` are indeed failing because their expectations are not the same after the change to the new undo manager. The default interval option for merging consecutive transactions is 1000 (1 second), and it appears all of them perform editing faster than this interval, then excepts undo to unapply a single change (which is not a correct expectation). For now, I set the default interval to be 0, and this will not merge any transaction, and will yield the expected result. I think maybe there is a need to take interval into consideration in future tests.

However, a new test has failed, and only when interval is 0. When set to 1000, that test didn't fail, but the previous 3 did. I think it has to do with some sort of event timing, and I'll look into it later on.

aaalsaleh:
@rrees It seems that if you change the intervals, then you pass/fail some tests. For example, once I set the default interval to 50, all the tests passed on my machine (last commit). Also, I tried to set the interval to 1000, and then test all the failing test manually by hand, and they all produced the expected output. Of course running it as an automated tests did not produce the expected output. I think there must be a race condition somewhere, and maybe it is related to one of the few setTimout callbacks?

rrees:
@aaalsaleh we'll need to get the tests passing reliably before we can merge the change. Feel free to change the tests to match the new behaviour but talking about a race condition makes me feel very nervous.

aaalsaleh:
@rrees @cutandpastey @hmgibson23 I've made major changes around the transaction manager, and now all the tests are passing.

The problems were:

1. `core/events` and `patch/events` had some issues differentiating between `setContent` and other mutations. Modifying all of them, and adding `forceMerge` option to `transactionManager.run` solved it.
2. `test/undo-manager.spec.js` was failing due to the assumption that two consecutive transactions should be considered two separate transactions. This is not the case with the new undo manager, especially with the default interval of 1000ms. Either change the tests to add some delay between the two typing events (> 1000ms), or what I did, change the interval to be 0 (no merge). This way, the two consecutive transactions will be considered two separate transactions.

I've also improved `pushHistory` by:

1. Saving more memory space by replacing last transaction's content when merging (instead of just pushing them together).
2. Checking now take into account the possibility of a shared undo manager. If the previous transaction was not from the same instance, or even was not a scribe transaction, the new transaction will not be considered a new one (not merged).

rrees:
Great, thanks. I'll check it out locally as soon as I can

aaalsaleh:
I've had to make some additional tweaks to handle edge cases. I've also add a new option `manager` under `undo` option. It can be used to pass a user defined undo manager. This will allow sharing a single undo manager instantiated by the user with multiple scribe instances (or other components), if needed. Of course, the shared undo manager must use the same undo manager API, or just reuse the one with scribe. I've quickly tested sharing a single undo manager with two scribe instances, and I can confirm it works as expected. Please let me know if you have a better alternative.

aaalsaleh:
More space/time efficiency improvements. **I'll no longer do any more commits, unless of course someone point a problem**. I've tested it with the automated tests with both Chrome and Firefox, and all tests were passed. One exception, is that `undo-manager.spec.js` with Firefox cannot accept options for scribe (Object Freezing as mentioned in the helpers source code). This is not the case with Chrome. Passing an the option `interval` of 0 is important for that test as I mentioned in another comment.

I've also manually tested more undo manager cases, and even put together a test example where two scribe instances share the same undo manager. There is a simple visualization of the undo stack in that example. Please check the `examples/shared-undo-manager.html`.

rrees:
@aaalsaleh You're right we should probably be merging rather than using defaults as regarding options. You're solution makes sense given what we do know.

I'm going to try and merge this in.

aaalsaleh:
@rrees Very glad to contribute something useful for such great project. Thanks for the opportunity.

rrees:
Sadly this didn't pass the tests for FF31 and 32 so we need to check how much we need to support those browsers

aaalsaleh:
Can you please list these tests, maybe it is something trivial to fix?

rrees:
Here's the [link to the failing build](https://travis-ci.org/guardian/scribe/builds/53633191). We are in the process of depreciating support for 31 and 32 so if you can fix it, great. Otherwise I think we will take this change once we've notified users of the change.

aaalsaleh:
If they are "950.1" and "950.2" in the link you provided, then it is just an issue with your helpers with Firefox. Passing the `options` parameter to `helpers.initializeScribe.bind` always freeze the test in Firefox. I've mentioned this with my last commit. It is a problem with the helpers, not the tests files or my commit. However, if it is another problem, I didn't see it in the link.

If you are willing to help me by modify the `scribe-test-harness/helpers.js` to add an additional helper function, then I fix the Firefox issue. Please add the following to that file:
```javascript
exports.setUndoInterval = function (interval) {
  return exports.driver.executeScript(function (interval) {
    window.scribe.options.undo.interval = interval;
  }, interval);
}
```
Once you do, I'll commit a change to `test/undo-manager.spec.js`, which will use this helper method to change the undo interval to be 0 before starting any of the undo manager tests.

I've tested this locally with firefox 31, and I can confirm it is passing with this fix. I can make a commit easily to `test/undo-manager.spec.js`, but not to `scribe-test-harness/helpers.js` since it is only a dev dependency.

rrees:
Hi there, I don't quite understand the issue. Ultimately we don't want options to changed after initialisation so I wouldn't want to add this but instead fix the initialisation issue.

cutandpastey:
@aaalsaleh you can do the above in the actual test without adding to the test harness.

aaalsaleh:
@cutandpastey haha, yes indeed, how did I miss that! I've just did that successfully.
@rrees no need, @cutandpastey was right, I already made the required change. Should I issue a new PR, or maybe reopen this one?

rrees:
Maybe submit a new PR and reference this one. Thanks!

PR #349, opened 2015-03-09
Testing: add Firefox 36

rrees:
The time is coming when we should bump the Firefox version number, this just adds 36 to the testing mix to see if there are any issues (it is currently passing)


cutandpastey:
:+1:

PR #351, opened 2015-03-09
Testing: add Chrome 40

rrees:
As we have auto-updating on this is now our number browser version so we can probably delete the others when this is merged.


cutandpastey:
:+1:

PR #352, opened 2015-03-09
Testing: remove old Chrome versions

rrees:
The current Chrome versions in use for the Guardian are 39, 40 and 41


cutandpastey:
:+1:

PR #355, opened 2015-03-11
Undo manager: test build

rrees:
This combines #353 with #346 and tweaks the interval of the undo slightly to have something that works a little better for fast typists.

### DIFF
--- a/src/scribe.js
+++ b/src/scribe.js
@@ -47,7 +47,12 @@ define([
     this.options = defaults(options || {}, {
       allowBlockElements: true,
       debug: false,
-      undo: { manager: null, enabled: true, limit: 100, interval: 1000 },
+      undo: {
+        manager: false,
+        enabled: true,
+        limit: 100,
+        interval: 250
+      },
       defaultCommandPatches: [
         'bold',
         'indent',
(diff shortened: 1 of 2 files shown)


aaalsaleh:
@rress, what you are trying is not necessary, nor going to make it pass. Please see my comments in #353 on the line 10 of `test/undo-manager.spec.js`. It was a problem with the order of the merge. Please remove the options passed in the `helpers.initializeScribe.bind`, it is the reason behind freezing. But also please keep the other lines added by #353 (we need to set the interval to 0, as the tests expect it to behave like that). Finally, a default interval of 250ms, is too low for slow typist to notice any difference. I was thinking of setting it to a default of 2000ms, since some people are not touch-typist.

rrees:
@aaalsaleh Regarding the interval I think there are two problems, firstly the first content item being empty makes an undo in the first line brutal (it just deletes the whole first line). Secondly, journalists are very fast typists so we need a default that works better for them and then allow people to override it for the rarer, slower case.

We'll be doing some QA here to try and get the balance right.

aaalsaleh:
@rrees Yes, I guess you are right about journalists. I was having in mind regular web users, since I come from a different background. However, keep in mind that fast typist rarely use undo to undo one or two words, but rather a bunch of words (probably lines), otherwise they would use backspace (or at least this is how I do it). For example, Microsoft Word undo several lines with no interval (only on paste/formatting), while OpenOffice undo on spaces and new lines (which is annoying). The behavior can certainly be improved further, but I guess passing the tests now is the first priority.

PR #356, opened 2015-03-13
Options: break out option handling into its own module

rrees:
With the addition of more options I thought it was worth breaking the option handling out to its own module which will also store the default option values. This module has its own unit-tests so that the functionality isn't being tested indirectly via things like the block handling tests.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@
 /examples/build.js
 .tern-port
 .npm-debug.log
+.DS_Store
--- a/src/config.js
+++ b/src/config.js
@@ -1,6 +1,6 @@
 define(['lodash-amd/modern/objects/defaults',], function (defaults) {
 
-	var defaultOptions = {
+    var defaultOptions = {
       allowBlockElements: true,
       debug: false,
       undo: {
--- a/src/config.js
+++ b/src/config.js
@@ -20,14 +20,14 @@ define(['lodash-amd/modern/objects/defaults',], function (defaults) {
     };
 
 
-	function checkOptions(userSuppliedOptions) {
-		var options = userSuppliedOptions || {};
+    function checkOptions(userSuppliedOptions) {
+        var options = userSuppliedOptions || {};
 
-		return Object.freeze(defaults(options, defaultOptions));
-	}
+        return Object.freeze(defaults(options, defaultOptions));
+    }
 
-	return {
-		defaultOptions: defaultOptions,
-		checkOptions: checkOptions
-	}
+    return {
+        defaultOptions: defaultOptions,
+        checkOptions: checkOptions
+    }
 });
--- a/src/config.js
+++ b/src/config.js
@@ -1,0 +1,33 @@
+define(['lodash-amd/modern/objects/defaults',], function (defaults) {
+
+	var defaultOptions = {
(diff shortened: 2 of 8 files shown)


src/config.js
```diff
@@ -0,0 +1,33 @@
...
+	function checkOptions(userSuppliedOptions) {
+		var options = userSuppliedOptions || {};
+
+		return Object.freeze(defaults(options, defaultOptions));
+	}
+
+	return {
+		defaultOptions: defaultOptions,
+		checkOptions: checkOptions
+	}
```
theefer: Indentation Gone Wild?

test/unit/config.spec.js
```diff
@@ -0,0 +1,32 @@
...
+		expect(checkedOptions.allowBlockElements).to.be.false;
+	});
+
+	describe('defaults', function() {
+		it('should apply default values', function() {
+			var checkedOptions = config.checkOptions({});
+
+			expect(checkedOptions.allowBlockElements).to.be.true;
+		});
+	});
```
theefer: Please use spaces to indent.

src/config.js
rrees: I clearly don't have editorconfig setup correctly...

test/unit/config.spec.js
rrees: This matches Editorconfig now

src/config.js
```diff
@@ -0,0 +1,33 @@
...
+      'outdent',
+      'createLink'
+    ]
+  };
+
+
+  function checkOptions(userSuppliedOptions) {
+    var options = userSuppliedOptions || {};
+
+    return Object.freeze(defaults(options, defaultOptions));
```
aaalsaleh: Correct me if I am wrong, but with this line you intend on freezing the options after scribe initialization. While this might be a good programming practice, I am afraid it will bring back the FF31/32 `test/undo-manager.spec.js` issue back.

hmgibson23: That's a good point. I think it will. But @rrees - are we retiring FF31/32 support?

hmgibson23:
Looks good other than the Object freeze thing (which I'm not sure about) :+1:

src/config.js
rrees: Well the tests still pass so I don't think we really understand the FF31/32 undo manager issue. But yes the plan is to depreciate support so at that point we won't need to try and re-write window.scribe.options (which we really don't want to happen).

test/unit/config.spec.js
theefer: Does it? I still see tabs here.

.gitignore
```diff
@@ -5,3 +5,4 @@
 /examples/build.js
 .tern-port
 .npm-debug.log
+.DS_Store
```
theefer: Put that in your global git config perhaps?

rrees: Good idea

test/unit/config.spec.js
rrees: I definitely have two space tabs locally and git is saying the branch is up to date. That said it does look wrong here.

theefer: I checked out the branch and there are definitely spaces. Are you sure you pushed all your changes?

PR #357, opened 2015-03-13
Build: Plumber: Upgrade 0.4

rrees:
Another attempt to upgrade Plumber now we've done some work on the test harness


theefer:
:+1: if it all works

PR #359, opened 2015-03-18
Added a null check in selection.js

shaunnetherby:
I'm running into a case in Shadow DOM where selection.anchorNode and selection.focusNode is null in Chrome, causing the following error in the console log every time the contenteditable div is selected:

Uncaught NotFoundError: Failed to execute 'setStart' on 'Range': The node provided was null

I can't figure out what is causing these nodes to be null and it doesn't seem to affect scribes functionality, so I just added an additional null check.

### DIFF
--- a/src/api/selection.js
+++ b/src/api/selection.js
@@ -24,7 +24,7 @@ function (elementHelper) {
       }
 
       this.selection = rootDoc.getSelection();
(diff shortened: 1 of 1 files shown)


rrees:
As long as the tests pass then I can't see the harm. I'll try and merge soon.

PR #360, opened 2015-03-19
Allows overriding of default formatters and plugins

DavidTobin:
This PR allows for overriding of default plugins and formatters. It extends the current behavior of allowing command patches to be overridden.

### DIFF
--- a/src/scribe.js
+++ b/src/scribe.js
@@ -1,13 +1,10 @@
 define([
+  'lodash-amd/modern/arrays/difference',
+  './plugins/core/plugins',
   './plugins/core/commands',
+  './plugins/core/formatters',
   './plugins/core/events',
-  './plugins/core/formatters/html/replace-nbsp-chars',
-  './plugins/core/formatters/html/enforce-p-elements',
-  './plugins/core/formatters/html/ensure-selectable-containers',
-  './plugins/core/formatters/plain-text/escape-html-characters',
-  './plugins/core/inline-elements-mode',
   './plugins/core/patches',
-  './plugins/core/set-root-p-element',
   './api',
   './transaction-manager',
   './undo-manager',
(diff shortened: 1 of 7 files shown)


src/scribe.js
```diff
@@ -109,27 +115,40 @@ define([
...
     if (this.allowsBlockElements()) {
-      // Commands assume block elements are allowed, so all we have to do is
-      // set the content.
-      // TODO: replace this by initial formatter application?
-      this.use(setRootPElement());
-      // Warning: enforcePElements must come before ensureSelectableContainers
-      this.use(enforcePElements());
-      this.use(ensureSelectableContainers());
+      /**
+       * Warning: enforcePElements must come before ensureSelectableContainers
```
rrees: I think we should keep this rule enforced in code

rrees:
This looks absolutely smashing, thanks!

Could I be a bit cheeky and ask whether I could merge #356 and then get you to add the defaults to that model? I was worried that main Scribe was getting a bit bloated.

DavidTobin:
Of course, work away!

src/scribe.js
DavidTobin: Will update!

DavidTobin:
@rrees Rebased with current master and made a change to force `setRootPElement` to load first. How does this look now?

rrees:
@DavidTobin I'm thinking this would be a great use case for ImmutableJS's OrderedSets!

In terms of this change though, could you extract the sort function into a named entity and add a unit-test for its behaviour?

I worried that the inline sort isn't that clear from a maintenance point of view.

Apart from that, good to go.

DavidTobin:
Makes total sense! I'll update soon.

DavidTobin:
Updated. Let me know if anything else is out of place.

rrees:
Okay so this is failing consistently on Firefox so I'll revert and we can try and figure out what is going on.

DavidTobin:
Looks to be a poor assertion in my override test.

rrees:
Can you raise a new pull request?

DavidTobin:
Will do.

PR #363, opened 2015-03-23
Allows overriding of default formatters and plugins

DavidTobin:
This PR allows for the default plugins and formatters to be overridden, so that users can include their own solutions for these.

### DIFF
--- a/src/config.js
+++ b/src/config.js
@@ -47,11 +47,11 @@ define([
 
     // Remove invalid plugins
     if (options.defaultPlugins) {
-      options.defaultPlugins    = (options.defaultPlugins || []).filter(filterByPluginExists(defaultOptions.defaultPlugins));
+      options.defaultPlugins    = options.defaultPlugins.filter(filterByPluginExists(defaultOptions.defaultPlugins));
     }
 
     if (options.defaultFormatters) {
-      options.defaultFormatters = (options.defaultFormatters || []).filter(filterByPluginExists(defaultOptions.defaultFormatters));
+      options.defaultFormatters = options.defaultFormatters.filter(filterByPluginExists(defaultOptions.defaultFormatters));
     }
 
     return Object.freeze(defaults(options, defaultOptions));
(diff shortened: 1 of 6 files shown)


DavidTobin:
Reference: https://github.com/guardian/scribe/pull/360

DavidTobin:
@rrees Here's the updated PR. The issue with the last one was that I wrongly assumed the same behavior across multiple browsers when the default plugins were disabled. I've updated the tests to contain no browser assumptions.

rrees:
Running the tests locally I am still getting consistent failures on Firefox. I can load the AMD example page though so I'm a bit confused about what is the problem here. Are you able to run the Firefox tests locally (I think I might need to [merge a change to scribe-test-harness](https://github.com/guardian/scribe-test-harness/pull/20))?

DavidTobin:
Ah, I think that might be it. We ran into some problems of our own and upgrading fixed for us. Firefox isn't running at all for me locally.

rrees:
Okay let me merge and release that other change and see what happens. I'll ping you with an update.

rrees:
If you merge master and `npm install` and then run `BROWSER_NAME=firefox ./run-tests.sh`

I'm hoping that you will get Firefox running and have some tests failing

DavidTobin:
Getting the failing tests now. Let me see if I can fix those. Will update soon.

DavidTobin:
Hrm, so the failing tests have begun to work for me out of seemingly nowhere. I've had 3 successful runs in a row now after constant failures. 

I'm working on a clean branch (`scribe-override-default-plugins`) that has been rebased with master. Is there a race condition that I am missing? Could you check Firefox on your end again also?

DavidTobin:
Okay, can reproduce the fails. If I give focus to another window while the tests are running on Firefox, the tests will fail. Let me see if I can come up with a solution.

DavidTobin:
Sorry for getting back to this so late. So it looks like the Firefox tests are working. Are you getting the same thing?

rrees:
Firefox is failing for me completely on this local branch now. I suspect that is probably due to something in my setup as it failing on master as well.

I'm going to merge and see how the Travis build goes.

rrees:
Victory! I'll do the release now

DavidTobin:
Great! :+1:

Issue #362, opened 2015-03-23
Tests break when including lodash filter in config.js


DavidTobin:
Adding `'lodash-amd/modern/collections/filter` in `config.js` breaks tests when running `./run-tests.sh`:

```
/home/david/code/scribe/bower_components/lodash-amd/modern/support.js:30
  support.funcDecomp = !isNative(window.WinRTError) && reThis.test(function() 
                                 ^
ReferenceError: window is not defined
    at Object.<anonymous> (/home/david/code/scribe/bower_components/lodash-amd/modern/support.js:30:34)
    at runFactory (/home/david/code/scribe/node_modules/node-amd-require/node_modules/amdefine/amdefine.js:181:30)
    at define (/home/david/code/scribe/node_modules/node-amd-require/node_modules/amdefine/amdefine.js:275:13)
    at Object.<anonymous> (/home/david/code/scribe/bower_components/lodash-amd/modern/support.js:9:1)
    at Module._compile (module.js:456:26)
    at Object.require.extensions..js (/home/david/code/scribe/node_modules/node-amd-require/index.js:59:15)
    at Module.load (module.js:356:32)
    at Function.Module._load (module.js:312:12)
    at Module.require (module.js:364:17)
    at req (/home/david/code/scribe/node_modules/node-amd-require/node_modules/amdefine/amdefine.js:142:31)
```

rrees:
Yeah, I've seen this too. I think this is a bug in the quite old version of Lodash we're using. I've been meaning to do a version upgrade.

PR #364, opened 2015-03-24
Dependencies: upgrade Lodash to 3.5.0

rrees:
Our current version of Lodash is quite old so this is partly a hygiene update but also it might address #362


cutandpastey:
:+1:

PR #373, opened 2015-04-01
Dependencies: update test harness to 0.0.20

rrees:
I want to try and use the generic spec discovery (or die trying)


cutandpastey:
:+1:

PR #375, opened 2015-04-01
Fix bug in EventEmitter.prototype.off

rf-:
The most recent version of this code was accidentally replacing the entire `this._listeners` object.

### DIFF
--- a/src/event-emitter.js
+++ b/src/event-emitter.js
@@ -19,9 +19,9 @@ define(['lodash-amd/modern/array/pull',
   EventEmitter.prototype.off = function (eventName, fn) {
     var listeners = this._listeners[eventName] || Immutable.Set();
     if (fn) {
-      this._listeners = listeners.delete(fn);
+      this._listeners[eventName] = listeners.delete(fn);
     } else {
-      this._listeners = listeners.clear();
+      this._listeners[eventName] = listeners.clear();
     }
   };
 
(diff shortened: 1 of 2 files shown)


rf-:
It looks like there's some kind of authentication problem with the build config for pull requests.

rrees:
Yes correct, it's an issue to do with [authentication in forks](https://github.com/travis-ci/travis-ci/issues/1946) as far as I can tell.

rrees:
Oooo, subtle, good spot. I'll merge and release tomorrow. We should also remove the if-else with short-circuiting.

theefer:
> We should also remove the if-else with short-circuiting.

I'm curious, why would you consider that better?

rrees:
It splits the code flow into the general case (adding and removing event handlers) and the special case (unbind all handlers for an event).

If you're into it it (and I am) it also reduces the [cyclomatic complexity](http://en.wikipedia.org/wiki/Cyclomatic_complexity) of the general branch. However for a single if-else this is pretty trivial gain so I'm not going to claim too much for it.

theefer:
I tend to consider early returns bad practice as it breaks referential transparency of a code block, but easier to talk about it when we're both back in, and focus on eating easter chocolate in the meantime. :-)

rrees:
I don't want to spend much more talking about it either (poor @rf- )! But the short version is that this function doesn't have referential transparency as it operates by side-effect (mutating `this`) so conventional code-pathing rules would apply.

theefer:
Fair enough in this example. 

I was more interested in your opinion in terms of general coding style guidelines, not in the context of this very helpful pull request!

PR #376, opened 2015-04-01
Fix focus handler when scribe.el is empty (fix #368)

### LINKED ISSUE #368: Focus handler breaks when Scribe's content has been set to ''
The [handler for the `focus` event](https://github.com/guardian/scribe/blob/master/src/plugins/core/events.js#L31) breaks when `scribe.el.firstChild` is `null`, which happens if the editor has been emptied using `scribe.setContent('')`. The specific error is `Uncaught TypeError: Failed to execute 'createTreeWalker' on 'Document': parameter 1 is not of type 'Node'.`.

I'm prepared to accept "don't do that" as an answer, but it'd be pretty inconvenient. It naively seems like `var focusElement = getFirstDeepestChild(scribe.el)` might fix it?

rf-:
These tests aren't very satisfying, but I wasn't sure if there was a better place to put them or a better approach for testing the bug. Let me know if you want me to change anything.

### DIFF
--- a/src/plugins/core/events.js
+++ b/src/plugins/core/events.js
@@ -23,12 +23,11 @@ define([
         // In Chrome, the range is not created on or before this event loop.
         // It doesn’t matter because this is a fix for Firefox.
         if (selection.range) {
-
           var isFirefoxBug = scribe.allowsBlockElements() &&
                   selection.range.startContainer === scribe.el;
 
           if (isFirefoxBug) {
-            var focusElement = getFirstDeepestChild(scribe.el.firstChild);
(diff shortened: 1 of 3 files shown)


rrees:
I think the easiest way to do this is to pull `getFirstDeepestChild` up to the api in module perhaps just called `children` for now. Rename the function to `firstDeepestChild` Write unit tests against the function for the normal case and the edge case you're fixing here, then retain the browser test that invokes the functionality but don't worry about trying to capture the exception.

rrees:
@rf- I've implemented a mix of your solution and my idea in a different PR #386

PR #377, opened 2015-04-01
Bump lodash dependency in package.json

rf-:
e7d3b00 bumped the lodash dependency in bower.json but not package.json.


forresto:
About to make the same PR.

rrees:
True dat

PR #379, opened 2015-04-08
Testing: discover spec files rather than specifying them

rrees:
This change should allow people to drop new spec files into the project and have them added to the test runner automatically.


cutandpastey:
:+1:

PR #380, opened 2015-04-09
Testing: update target SauceLabs browsers

rrees:
The new minimum supported versions are FF 36 and Chrome 4.

Updates the Readme and the SauceLabs matrix


hmgibson23:
:+1:

PR #382, opened 2015-04-24
Observing 'attributes' is unnecessary

bradvogel:
Observing 'attributes' is actually unnecessary since all mutations that don't add or remove DOM nodes are filtered out anyways. See `includeRealMutations`. This will increase performance a bit.

### DIFF
--- a/src/dom-observer.js
+++ b/src/dom-observer.js
@@ -50,7 +50,6 @@ define([
     });
 
     observer.observe(el, {
-      attributes: true,
       childList: true,
       subtree: true
     });


hmgibson23:
Makes sense to me :+1:

bradvogel:
The build is failing. Is that expected?

bradvogel:
Bump @hmgibson23

hmgibson23:
Yeah build always fails from foreign PRs ... sorry this slipped away from me. I'll check and merge today.

PR #385, opened 2015-05-18
Edit insertHTML fix: don't purge user-generated SPANs

christopherliu:
A while back, this was written in insert-html.js: "FIXME: what if the user actually wants to use SPANs? This could cause conflicts."
Turns out that user was me. This updated patch is more targeted, removing only SPANs that are truly superfluous. In addition, it also covers more than one SPAN, where the original code would only cover the first child SPAN.

This was originally #339


christopherliu:
Thanks!

PR #386, opened 2015-05-29
Children search: return the root node for empty nodes

rrees:
This is my proposed fix to #368 that builds on the work of #376 but adds a small separation of code and a unit-test for the case that the Scribe content is empty.

I decided not to use a mock browser for the document substitution so this still isn't very testable for genuine tree traversal but it should address people's immediate problems.

### DIFF
--- a/src/api/children.js
+++ b/src/api/children.js
@@ -1,7 +1,11 @@
 define(function () {
 
   'use strict';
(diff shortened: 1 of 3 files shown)


hmgibson23:
Looks good :+1:

Issue #388, opened 2015-06-10
Some contributions to Scribe


regiskuckaertz:
Hello, a few weeks ago I started using Scribe for an internal project here at the Government of Luxembourg. It's a really great piece of software with an already comprehensive feature set, which is quite impressive given its short lifespan so far. As I was getting acquainted with how the code's organized, I thought about a few improvements (mostly moving things around) and committed them to [my own fork](https://github.com/regiskuckaertz/scribe). Here they are, for the most part:

- moved all DOM-related operations inside ./node.js (no more ./element.js, ./api/children.js or ./api/node.js), plus a few new ones to encapsulate some recurring operations in plugins
- removed dependancy on lodash (except in ./config.js)
- used the awesome Node.compareDocumentPosition and Node.contains everywhere I could as they're super fast
- rewrote algorithms using the TreeWalker API with the Node/Selectors API instead as it's usually much faster
- removed temporary variables where I could
- moved code around to spare the JS engine of redundant calculations
- used Immutable instead of traditional arrays here and there (see wrapChildNodes and UndoManager for instance)—structural sharing FTW!

I still have to do exhaustive tests, and there are areas I need to wrap my head around (pushHistory in particular), but in the meantime if you find anything worth adding to the master repo, please tell me and I'll be happy to contribute—I'm not a GitHub pro, so I might need some guidance here.

I also think Immutable could be used in many other places, so I will keep working on that, but as most of the code relies on vars and procedural code, I'm thinking of using transducers instead—I will definitely have look at that as well and keep you posted.

hmgibson23:
Hi,
Glad that you've been having success using scribe. I think the least controversial and perhaps most helpful to us would the speed improvement stuff (using the TreeWalker everywhere and the Node.compareDocumentPosition stuff). Removal of temporary variables sounds good too. 

I'd like to see lodash go, as personally I have my doubts about it's effectiveness for this type of project but other contributors might have different opinions. 

If you want to make the pull request with just the stuff I mentioned I'll be happy to review it and get it in the repo.

Thanks!

regiskuckaertz:
@hmgibson23 awesome! I will isolate all these things and make a PR asap.

JeSuis:
:clap:

regiskuckaertz:
Closing this issue as we've moved on with the PRs.

PR #389, opened 2015-06-11
Some updates for increased efficiency

regiskuckaertz:
Hi,
As discussed in [#388](https://github.com/guardian/scribe/issues/388), I removed anything that was not strictly related to speeding up the execution of Scribe, primarily by removing TreeWalker which is slow as hell. I hope everything's in order for proper testing; I'm eager to help so please don't hesitate.


src/api/children.js
```diff
@@ -7,19 +7,12 @@ define(function () {
...
-        return firstDeepestChild(treeWalker.currentNode);
-      }
-    } else {
-      return treeWalker.currentNode;
+    var child = node.firstChild;
+    if( child.nodeName === 'BR' ) {
+      return node;
     }
+
+    return firstDeepestChild(child);
```
hmgibson23: Realise this isn't related but I'm not convinced a recursive solution is the most efficient? Maybe a NodeIterator would be better.

regiskuckaertz: As far as my tests are concerned, NodeIterator and TreeWalker both perform really slow compared to Node or Selectors, that's why I ruled them out. But you are right, recursion might be slower than a loop—I don't know if browser engines optimize tail-recursive calls. At this point it might actually depend on the engine: http://jsperf.com/tail-recursive-vs-loop

regiskuckaertz:
oops, if I read it correctly, all failed tests seem to lead to a line-height and/or a misplaced BR. I'll check that out.

src/api/node.js
```diff
@@ -10,26 +10,23 @@ define(function () {
...
       return;
     }
 
-    var currentNode = this.node.parentNode;
+    var currentNode = this.node;
 
     // If it's a `contenteditable` then it's likely going to be the Scribe
     // instance, so stop traversing there.
-    while (currentNode && ! isTopContainerElement(currentNode)) {
+    while ((currentNode = currentNode.parentNode) && currentNode !== rootElement) {
```
hmgibson23: Not quite sure how this is more efficient? Other than the function call ... but that seems pretty negligible.

src/api/selection.js
```diff
@@ -1,202 +1,118 @@
...
-        return;
-      }
+      if (!range) { return; }
 
-      //we need to ensure that the scribe's element lives within the current document to avoid errors with the range comparison (see below)
-      //one way to do this is to check if it's visible (is this the best way?).
-      if (!scribe.el.offsetParent) {
+      //we need to ensure that the scribe's element lives within the current document
+      //to avoid errors with the range comparison (see below)
+      if (!document.contains(scribe.el)) {
```
hmgibson23: Not sure if this is what we want. How will this work with multiple scribe elements? It should be okay but a test might be nice.

src/dom-observer.js
```diff
@@ -1,30 +1,22 @@
...
-) {
+], function (elementHelpers, nodeHelpers) {
+
+  function includeRealMutations(mutations) {
+    var realChangedNodes = mutations
+      .map(function(mutation) {
+        return mutations.slice.call(mutation.addedNodes)
+          .concat(mutations.slice.call(mutation.removedNodes));
+      })
+      .reduce(function(result, input) { return result.concat(input); }, [])
```
hmgibson23: This is nice!

src/api/node.js
regiskuckaertz: 100% agreed, what I intended is to get rid of the func var, so using `function isTopContainerElement()` instead of `var isTopContainerElement = function()` would do it as well. Do you prefer it that way?

regiskuckaertz: (the benefit is that the function is defined at parse-time instead of each time `getAncestor` is called)

hmgibson23: Okay that makes sense. Scribe tends to prefer ```function``` declarations over ```var x = function``` stuff. It's not too unclear as it is so I'm happy for it to stay this way.

src/api/selection.js
regiskuckaertz: I'm not sure I understand: you mean if there are two scribe instances in the same document? The thing with `offsetParent` is that is null either when the node is attached to a `DocumentFragment` or when its `display` style property (or one of its ancestors') is set to `none`.

hmgibson23: Okay that should be fine then. Just wanted to check.

src/api/node.js
regiskuckaertz: thanks :+1:

src/dom-observer.js
regiskuckaertz: Thanks :-)

hmgibson23:
I'm happy with this. Be good to get a second set of eyes @rrees @cutandpastey

hmgibson23:
There's a few style issues. I won't go over them all but loosely, we use 2 spaces for basic indent and wrap all expressions with curly braces. So ```if``` etc. should always be accompanied by curly braces

cutandpastey:
Sadly, I have a whole bunch of failing tests.

regiskuckaertz:
@cutandpastey Hello, yes I should have done all the testing prior to doing the PR, but I had an error while running `setup.sh` with the certificate of the selenium domain name. As soon as I fix this, I'll be able to debug all my mistakes. I'll keep you posted asap. (Also, I've been profiling memory consumption and you should expect another PR with *named* function, it's much easier to debug :)

regiskuckaertz:
@hmgibson23 Alright, I will go over all my changes and fix coding style issues. Yeah!

src/api/selection.js
rrees: Do you want the negation there?

src/api/selection.js
```diff
@@ -223,13 +138,9 @@ function (elementHelper) {
...
-      if (markers.length >= 2) {
-        newRange.setEndAfter(markers[1]);
-      } else {
-        // We always reset the end marker because otherwise it will just
-        // use the current range’s end marker.
-        newRange.setEndAfter(markers[0]);
-      }
+      // We always reset the end marker because otherwise it will just
+      // use the current range’s end marker.
+      newRange.setEndAfter(markers[1] || markers[0]);
```
rrees: I think is quite confusing, could you inline a ternary operator here to have something a little clearer?

regiskuckaertz: @rrees Absolutely, yes.

src/event-emitter.js
```diff
@@ -28,15 +29,15 @@ define(['lodash-amd/modern/array/pull',
...
-    var events = eventName.split(':');
-    while(!!events.length){
-      var currentEvent = events.join(':');
-      var listeners = this._listeners[currentEvent] || Immutable.Set();
-      //trigger handles
-      listeners.forEach(function (listener) {
-        listener.apply(null, args);
-      });
-      events.splice((events.length - 1), 1);
+    eventName = eventName.split(':');
```
rrees: Don't rebind the function parameters. That's really confusing.

src/api/selection.js
regiskuckaertz: @rrees I think yes, as per the comment just above: if the element is not inside the document, then quit.

src/event-emitter.js
regiskuckaertz: I may go overboard sometimes when it comes to stack usage. This change will be reverted tonight. Thanks for your feedback @rrees !

rrees:
It's great that you've given so much love to the performance! It's been at the back of my mind for ages. If you can get the tests to pass then let's merge. If not do you want to try and break this PR into smaller parts so we can start merging the simpler parts.

regiskuckaertz:
@rrees thanks a lot Robert, I appreciate. I have many ideas on how to improve the performance of Scribe. This is the first time I do such a huge PR, and it taught me that it's probably not the best way to help you, given how much work you have to do afterwards. Let's break the PR apart so that I can work on the failing tests while you can move forward, shall we? (I may need some guidance on how to do that with Git)

rrees:
@regiskuckaertz The simplest way to do it is to use branches

Probably the easiest way to do this now is to run the following, which will reset your master to Scribe's master and save all the current changes you've got to :

```
git remote add upstream https://github.com/guardian/scribe.git
git branch all-changes
git fetch upstream
git reset --hard upstream/master
```

You can now branch and make a smaller change and submit that via the Github interface:

```
git checkout -b my-small-change
git push origin my-small-change
```

regiskuckaertz:
Closing this one for good.

PR #393, opened 2015-06-16
Replaced TreeWalker calls with basic Node navigation

regiskuckaertz:
And here is a faster implementation of firstDeepestChild

### DIFF
--- a/src/api/children.js
+++ b/src/api/children.js
@@ -7,19 +7,12 @@ define(function () {
       return node;
     }
 
-    var treeWalker = document.createTreeWalker(node, NodeFilter.SHOW_ALL, null, false);
-    var previousNode = treeWalker.currentNode;
-    if (treeWalker.firstChild()) {
-      // TODO: build list of non-empty elements (used elsewhere)
-      // Do not include non-empty elements
-      if (treeWalker.currentNode.nodeName === 'BR') {
-        return previousNode;
-      } else {
-        return firstDeepestChild(treeWalker.currentNode);
-      }
-    } else {
-      return treeWalker.currentNode;
+    var child = node.firstChild;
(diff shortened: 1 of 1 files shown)


hmgibson23:
:+1: for smaller PRs

src/api/children.js
```diff
@@ -7,19 +7,12 @@ define(function () {
...
-        return firstDeepestChild(treeWalker.currentNode);
-      }
-    } else {
-      return treeWalker.currentNode;
+    var child = node.firstChild;
+    if( child.nodeName === 'BR' ) {
+      return node;
     }
+
+    return firstDeepestChild(child);
```
hmgibson23: I don't see how recursion is more efficient than the Treewalker solution I'm afraid.

hmgibson23: Okay actually I misread the diff ... this is fine but it would be far more efficient to use the TreeWalker all the way down rather than recurring.

hmgibson23: You should just be able to TreeWalk all the way to the bottom and then just get the last node. https://developer.mozilla.org/en/docs/Web/API/TreeWalker

regiskuckaertz: I think where TreeWalker really shines is where you have a complex DOM and only want to browser through specific nodes. In this case, where only `firstChild` is accessed, the cost of creating the TW is just too much. See a similar test case here: https://jsperf.com/treewalker-vs-dom-traversal or my proposal for a faster implementation of `isEmptyInlineElement` https://jsperf.com/node-vs-treewalker-apis

rrees: Okay, I find the Perf tests convincing.

rrees:
:+1:

PR #395, opened 2015-06-16
Faster selection api

regiskuckaertz:
* Removed unnecessary dependency on AMD module `element` (already available in `scribe`)
* Moved the `rootDoc` selection out of the constructor so that it is executed only once (per Scribe instance)
* Used `compareDocumentPosition` to detect if `scribe.el` is in a fragment, and only then look for that fragment
* Moved redundant code in small, optimizable functions `createMaker` and `insertMarker`
* Replaced range arithmetic with simpler node comparisons
* Replaced `Element.offsetParent` test with `document.contains`, more accurate
* Overall simpler `placeMarker` implementation will less transient variables
* Faster implementation of `isEmtpyInlineElement` (moved to node.js for consistency)
* Moved `scribe.node` and `scribe.element` up in the prototype chain so they're assigned at parse time

### DIFF
--- a/src/api/selection.js
+++ b/src/api/selection.js
@@ -104,14 +104,14 @@ define(function () {
       //if this isn't true scribe will place markers within the selections parent
       //we want to ensure that scribe ONLY places markers within it's own element
       if (scribe.el.contains(range.startContainer) && scribe.el.contains(range.endContainer)) {
-        // insert start marker
-        insertMarker(range.cloneRange(), createMarker());
+        // End marker
+        var rangeEnd = range.cloneRange();
+        rangeEnd.collapse(false);
+        insertMarker(rangeEnd, createMarker());
 
         if (! range.collapsed ) {
-          // End marker
-          var rangeEnd = range.cloneRange();
-          rangeEnd.collapse(false);
-          insertMarker(rangeEnd, createMarker());
+          // insert start marker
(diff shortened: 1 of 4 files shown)


rrees:
Looks good to me but I think I'd like a second-pair of eyes :+1:

src/api/selection.js
```diff
@@ -1,44 +1,76 @@
...
-
-        this.range.setStart(this.selection.anchorNode, this.selection.anchorOffset);
-        reverseRange.setStart(this.selection.focusNode, this.selection.focusOffset);
 
-        // Check if anchorNode is before focusNode, use reverseRange if not
-        if (this.range.compareBoundaryPoints(Range.START_TO_START, reverseRange) <= 0) {
+        // Check if anchorNode is before focusNode, reverse the range if not
+        if (this.selection.anchorNode === this.selection.focusNode) {
+          this.range.setStart(this.selection.anchorNode, Math.min(this.selection.anchorOffset, this.selection.focusOffset));
+          this.range.setEnd(this.selection.anchorNode, Math.max(this.selection.anchorOffset, this.selection.focusOffset));
```
hmgibson23: Can you abstract this a little more? It's quite confusing in its current state.

regiskuckaertz: Sure!

hmgibson23:
Looks good :+1:

PR #396, opened 2015-06-16
Faster enforce p element

regiskuckaertz:
* Removed dependency on lodash
* Moved `wrapChildNodes` and `traverse` inside the closure to avoid passing around `scribe`
* Rewrote `traverse` without TreeWalker and recursive calls
* Wrote a more efficient `wrapChildNodes` using immutable data structures

### DIFF
--- a/src/plugins/core/formatters/html/enforce-p-elements.js
+++ b/src/plugins/core/formatters/html/enforce-p-elements.js
@@ -1,8 +1,6 @@
 define([
-  'lodash-amd/modern/array/last'
-], function (
-  last
-) {
+  'immutable/dist/immutable'
+], function (Immutable) {
 
   /**
    * Chrome and Firefox: Upon pressing backspace inside of a P, the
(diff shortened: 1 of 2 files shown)


rrees:
Looks good to me, can I get a second opinion? :+1:

src/plugins/core/formatters/html/enforce-p-elements.js
```diff
@@ -21,69 +19,41 @@ define([
...
+      function wrapChildNodes(parentNode) {
+        var index = 0;
+        Immutable.List(parentNode.childNodes)
+          .filter(function(node) {
+            return node.nodeType === Node.TEXT_NODE || !elementHelpers.isBlockElement(node);
+          })
+          .groupBy(function(node, key, list) {
+            return key === 0 || node.previousSibling === list.get(key - 1) ?
+              index :
+              ++index;
```
hmgibson23: Can you use postfix notation here, prefix will not make any difference here speedwise as the compile will just optimise it away?

src/plugins/core/formatters/html/enforce-p-elements.js
```diff
@@ -21,69 +19,41 @@ define([
...
-    // FIXME: does this recurse down?
-
-    while (node) {
-      // TODO: At the moment we only support BLOCKQUOTEs. See failing
-      // tests.
-      if (node.nodeName === 'BLOCKQUOTE' && ! node._isWrapped) {
-        wrapChildNodes(scribe, node);
-        traverse(scribe, parentNode);
-        break;
+        while (node = parentNode.children[++i]) {
```
hmgibson23: Same here.

src/plugins/core/formatters/html/enforce-p-elements.js
```diff
@@ -21,69 +19,41 @@ define([
...
+          if( node.tagName === 'BLOCKQUOTE' && ! node._isWrapped ) {
```
hmgibson23: We don't usually put spaces round our if statements.

src/plugins/core/formatters/html/enforce-p-elements.js
regiskuckaertz: Yes, of course. BTW, is `_isWrapped` still necessary? It looks like a necessity of the previous implementation

hmgibson23: Of the top of my head I don't know. Try removing it and running the tests?

rrees: If each node is only visited once then it isn't necessary any more

PR #398, opened 2015-06-18
Change from Lodash to Immutable

rrees:
Inspired by the wave of recent changes from @regiskuckaertz I thought I'd try and clean up some of the remaining dependencies on Lodash and also constant declarations that we make that are not explicitly Immutable.

### DIFF
--- a/src/element.js
+++ b/src/element.js
@@ -2,11 +2,11 @@ define(['immutable/dist/immutable'], function (Immutable) {
 
   'use strict';
 
-  var blockElementNames = Immutable.fromJS(['ADDRESS', 'ARTICLE', 'ASIDE', 'AUDIO', 'BLOCKQUOTE', 'CANVAS', 'DD',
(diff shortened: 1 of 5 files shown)


src/element.js
```diff
@@ -1,14 +1,14 @@
-define(['lodash-amd/modern/collection/contains'], function (contains) {
+define(['immutable/dist/immutable'], function (Immutable) {
```
cutandpastey: Can we please reference in the require config this like the other dependencies?

regiskuckaertz:
Awesome :+1: I have a rewrite of UndoManager with Immutable, let me know if it can help

cutandpastey:
Have spoken to @rrees, we will alias Immutable in a separate PR. :+1:

src/element.js
theefer: Good call

PR #399, opened 2015-06-18
Require: Path changes for Immutable

rrees:
As requested by @cutandpastey in #398 the Immutable alias has now been changed to not include the path in the define statements


cutandpastey:
:+1:

PR #403, opened 2015-06-24
Replaced Array.map+slice+reduce+filter with Array.some

regiskuckaertz:
A small change which could potentially yield great results (think cut-and-paste large documents). In `includeRealMutations`, we just want to return as soon as a matching node is found, hence `Array.some`.

### DIFF
--- a/src/dom-observer.js
+++ b/src/dom-observer.js
@@ -14,8 +14,8 @@ define([
 
   function includeRealMutations(mutations) {
     return mutations.some(function(mutation) {
-      return mutations.some.call(mutation.addedNodes, hasRealMutation) ||
-        mutations.some.call(mutation.removedNodes, hasRealMutation);
+      return Array.prototype.some.call(mutation.addedNodes, hasRealMutation) ||
+        Array.prototype.some.call(mutation.removedNodes, hasRealMutation);
     });
   }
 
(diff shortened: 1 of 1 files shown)


src/dom-observer.js
```diff
@@ -3,24 +3,23 @@ define([
...
+    window.MozMutationObserver;
+
+  function hasRealMutation(n) {
+    return ! nodeHelpers.isEmptyTextNode(n) &&
+      ! elementHelpers.isSelectionMarkerNode(n);
+  }
 
+  function includeRealMutations(mutations) {
+    return mutations.some(function(mutation) {
+      return mutations.some.call(mutation.addedNodes, hasRealMutation) ||
```
rrees: Should the inner some actually be a prototype call since you're passing in the new data set to evaluate?

regiskuckaertz: Yes, in this case `mutation**s**.some` is `Array.prototype.some`, it just saved me a few keystrokes :-)

rrees: Can you put the full version to make it easier to read/maintain? It threw me at first.

regiskuckaertz: Of course! Done.

rrees:
:+1:

Issue #397, opened 2015-06-17
Consolidating node operations


regiskuckaertz:
Hello,

Right now DOM-related operations are scattered all over: in ./node.js, ./element.js, ./api/node.js, ./api/children.js and sometimes in core plugins (think `unwrap`, `insertAfter`, `hasContent`, etc.). Plus ./api/node.js defines a class with a link to a DOM node: this can be a source of memory leaks if objects are never released (in closures for instance).

I could survey the code, make an inventory of all DOM-related operations and put everything into a single AMD module (./node.js). This would increase consistency, remove redundant code, which means more optimized code and an overall happier browser :-)

How does that sound?

hmgibson23:
Sounds good to me!

regiskuckaertz:
@hmgibson23 great! I'm on it.

regiskuckaertz:
This is going to be a biggie, I'll split this into multiple PRs for everyone's sake :-)

rrees:
Are we okay to close this one now?

regiskuckaertz:
Yes of course.

PR #404, opened 2015-06-24
Moved all DOM helpers in src/node.js

regiskuckaertz:
In reference to https://github.com/guardian/scribe/issues/397

This is a first step, I spotted a lot of redundant code that could be replaced with calls to DOM helpers. I'll leave that to another PR not to overwhelm you guys.


src/node.js
```diff
@@ -34,6 +67,35 @@ define(function () {
...
+      currentNode = currentNode.parentNode;
+    }
+  }
+
+  function nextAll(node) {
+    var all = [];
+    while (node = node.nextSibling) {
+      all.push(node);
+    }
+    return all;
```
rrees: Can we make this return an immutable collection?

src/node.js
```diff
@@ -34,6 +67,35 @@ define(function () {
...
+    // instance, so stop traversing there.
+    while (currentNode && ! isTopContainerElement(currentNode)) {
+      if (nodeFilter(currentNode)) {
+        return currentNode;
+      }
+      currentNode = currentNode.parentNode;
+    }
+  }
+
+  function nextAll(node) {
```
rrees: Maybe call this allFollowingNodes? The name isn't quite right.

rrees:
Looks like a sensible consolidation :+1: 

I wouldn't want to see files of more than 200 or 300 lines though.

regiskuckaertz:
I agree 100%, at some point this module can be split in sub-modules, similar to api.js.

src/node.js
regiskuckaertz: Sure!

src/node.js
regiskuckaertz: Oh, that's a good idea, I'm on it.

src/node.js
regiskuckaertz: What do you think of nextSiblings? More idiomatic, though using inflections might throw people off ("I did not see the s").

rrees: `nextSiblings` sounds a good choice for now

rrees:
Can you merge in the upstream master? Then I'll try and release next week.

Issue #401, opened 2015-06-23
Clipboard paste


alexeygolev:
Getting "event.clipboardData.types.contains is not a function" when attempting to paste. The same here http://guardian.github.io/scribe/. Chrome 43.0.2357.130,  45.0.2439.0 canary, Opera 30.0.1835.59, Safari 8.0.6. Works in Firefox though

hntrdglss:
I'm experiencing the same, using the NPM release with webpack.

rrees:
@alexeygolev I've deployed 1.4.6 to the Github pages, can you retest?

rrees:
@alexeygolev @cutandpastey @regiskuckaertz Regis pointed out that FF might not be Array-like so the fix may have broken FF. Can you check?

We might need to use a fixed data structure to reconcile the behaviours (at least I understand why the Lodash function was being used now).

alexeygolev:
@rrees it indeed did break in FF. Works in Chrome, Opera and Safari

rrees:
@alexeygolev Okay, I have taken on your suggestion and this should now work for all browsers (release 1.4.8 or later).

rrees:
@alexeygolev I think release 1.4.13 might be the definitive solution for this. Do you want re-test and confirm?

alexeygolev:
@rrees yep... works across the browsers...cheers

PR #405, opened 2015-06-25
Paste: Unify clipboard types data types

rrees:
Combines @alexygolev 's [advice](https://github.com/guardian/scribe/issues/401#issuecomment-115260565) with my own desire to hit everything with the Immutable hammer.

Should finally solve #401 

An illustration of the unification: https://jsfiddle.net/heopbyo3/3/

### DIFF
--- a/src/plugins/core/events.js
+++ b/src/plugins/core/events.js
@@ -1,9 +1,11 @@
 define([
   '../../dom-observer',
-  '../../api/children'
+  '../../api/children',
+  'immutable'
 ], function (
   observeDomChanges,
-  children
+  children,
+  Immutable
 ) {
 
   'use strict';
@@ -184,7 +186,7 @@ define([
         if (event.clipboardData) {
           event.preventDefault();
 
-          if (event.clipboardData.types.indexOf('text/html') !== -1) {
+          if (Immutable.List(Array.from(event.clipboardData.types)).includes('text/html')) {
             scribe.insertHTML(event.clipboardData.getData('text/html'));
           } else {
             scribe.insertPlainText(event.clipboardData.getData('text/plain'));


regiskuckaertz:
:+1:

regiskuckaertz:
One never gets back from Immutableland :-)

alexeygolev:
@rrees Maybe a bit too late, but I checked the implementation of the Immutable.List constructor and figured that you can actually just use `var types = Immutable.List(event.clipboardData.types);`
It's possible because to create a List it goes through a sequence of checks, one of which is `isArrayLike`. In all the browsers our type thingy is an array like as far as Immutable concerned (it has a length method on its prototype) so Immutable calls the ArraySeq constructor as a result. That doesn't solve our problem on it's own because we just basically packed our array-like into a ArraySeq container. However, the only method we need is `.includes` which doesn't care about the kind of iterable you feed it as long as it can iterate over it using a for loop.
I didn't think about it because I completely forgot that we already have a dependency on Immutable.

rrees:
@alexeygolev Thanks for looking for this. I tried this but it failed for me on Firefox because the DOMStringList couldn't be tickled into the a Seq. If you're certain it works I would much prefer to use this as the solution.

alexeygolev:
@rrees weird... it shouldn't even depend on the FF version. Does this work for you in FF? https://jsfiddle.net/heopbyo3/4/

rrees:
That does appear to work, how odd that I thought it didn't. I'll try and release this (nicer) version soon.

rrees:
@alexeygolev unless you want to put in a PR?

alexeygolev:
@rrees will do it now

PR #406, opened 2015-06-26
Catch unexpected error when calling queryCommandState in Firefox

### LINKED ISSUE #208: Strange NS_ERROR_UNEXPECTED in CommandPatch.prototype.queryState
I wrote this to the maillist, but decided to make it into an actual issue since I've narrowed the problem a bit more down.

I'm using Firefox 30.0 on Mac OS X, and sometimes I get this strange error in the console: `NS_ERROR_UNEXPECTED:` and nothing more.

It says that it happens at line 3669 in scribe.js:

```javascript
CommandPatch.prototype.queryState = function () {
  return document.queryCommandState(this.commandName);  // line 3669
};
```

I've figured out that the error can be "omitted" with the following:

```javascript
CommandPatch.prototype.queryState = function () {
  if( this.commandName == 'insertUnorderedList' || this.commandName == 'insertOrderedList') {
      return false
  }
  return document.queryCommandState(this.commandName);
};
```

So somehow the list commands are forcing Firefox to throw an `NS_ERROR_UNEXPECTED`. Sometimes it's even `NS_ERROR_FAILURE`. But both cases have no more information associated.

Refrence to errors in Firefox: https://developer.mozilla.org/en-US/docs/Table_Of_Errors

jamespamplin:
Firefox will consistently throw an error when calling `queryCommandState` for `insertUnorderedList` and `insertOrderedList` commands when there's no selection or a contenteditable hasn't got focus. Error as follows:
```
[Exception... "Unexpected error"  nsresult: "0x8000ffff (NS_ERROR_UNEXPECTED)"  location: "JS frame :: debugger eval code :: <TOP_LEVEL> :: line 1"  data: no]
```

`scribe-plugin-toolbar` uses this command to determine the state of the list buttons. The failed call would result in the list command buttons still remaining active when scribe loses focus.

This issue has been recurring for a while in Firefox builds, so it makes sense to catch this until they get around to fixing it, see: https://bugzilla.mozilla.org/show_bug.cgi?id=562623

Fixes #208.

Also, added ability to pass an argument to `run-tests.sh` so you can specify which test specs to run.

### DIFF
--- a/src/plugins/core/patches/commands/insert-list.js
+++ b/src/plugins/core/patches/commands/insert-list.js
@@ -109,6 +109,19 @@ define([], function () {
         }.bind(this));
       };
 
+      // Explicitly catch unexpected error when calling queryState - bug in Firefox: https://github.com/guardian/scribe/issues/208
+      InsertListCommandPatch.prototype.queryState = function() {
+        try {
+          return scribe.api.CommandPatch.prototype.queryState.apply(this, arguments);
+        } catch (err) {
+          if (err.name == 'NS_ERROR_UNEXPECTED') {
+            return false;
+          } else {
+            throw err;
+          }
+        }
+      };
+
       scribe.commandPatches.insertOrderedList = new InsertListCommandPatch('insertOrderedList');
       scribe.commandPatches.insertUnorderedList = new InsertListCommandPatch('insertUnorderedList');
     };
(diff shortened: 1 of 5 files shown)


src/plugins/core/patches/commands/insert-list.js
```diff
@@ -109,6 +109,19 @@ define([], function () {
         }.bind(this));
       };
 
+      // Explicitly catch unexpected error when calling queryState - bug in Firefox: https://github.com/guardian/scribe/issues/208
```
rrees: Maybe put the comment above the if statement?

rrees:
That's rough! :+1:

PR #407, opened 2015-06-28
Updated UndoManager with immutable data structures

regiskuckaertz:


hmgibson23:
Looks good to me :+1: and nice to see more use of ImmutableJS!

src/undo-manager.js
```diff
@@ -18,70 +20,80 @@ define(function () {
...
-      }
-      this.position--;
+    if (this.position === 0) { return; }
 
-      if (this._fireEvent) {
-        this._ush.dispatchEvent(new CustomEvent('redo', {detail: {transactions: this._stack[this.position].slice()}, bubbles: true, cancelable: false}));
-      }
+    this.position--;
+    var transactions = this._stack.get(this.position);
+    var i = -1;
```
rrees: I think an explicit for loop would be clearer here rather than the while with the external loop variable.

src/undo-manager.js
```diff
@@ -18,70 +20,80 @@ define(function () {
...
   };
 
   UndoManager.prototype.item = function (index) {
-    if (index >= 0 && index < this.length) {
-      return this._stack[index].slice();
-    }
-    return null;
+    return index >= 0 && index < this.length ?
+      this._stack.get(index).toArray() :
+      null;
```
rrees: I know this is what the code used to do but would it be better if it was `undefined` that was returned at this point?

rrees:
Looks good but as this is an important feature we might need to spend some time doing a bit of manual QA.

src/undo-manager.js
regiskuckaertz: Yep sure.

src/undo-manager.js
regiskuckaertz: in fact, `null` is the correct value according to this: https://dvcs.w3.org/hg/undomanager/raw-file/tip/undomanager.html#dom-undomanager-item

regiskuckaertz:
Of course yes! I was wondering, do you have some kind of roadmap or a list of bugs you'd like to fix? I could be of much better help doing that.

src/undo-manager.js
rrees: I stand corrected!

rrees:
@regiskuckaertz We use the [Available](https://github.com/guardian/scribe/labels/available) label to indicate things that people could tackle.

In some cases though these issues are quite old so we might need to discuss whether the issue is still relevant or not.

regiskuckaertz:
@rrees oh there is work :-) I'll see what I can do.

PR #410, opened 2015-06-29
Restore scribe.element

rrees:
Alias scribe.element to scribe.node for the benefit of plugins and maintain backwards compatibility (introduced in #404)


hmgibson23:
:+1:

PR #412, opened 2015-06-29
chore(events): Removed redundant slice of clipboardData.types

alexeygolev:
Simplifying #405 fix for #401. That should work across all browsers. I'll keep an eye on any issues popping here that might be connected.


alexeygolev:
Travis CI fails because of some authentication issues. No idea why

hmgibson23:
@alexeygolev - it's because of a foreign creds issue. Nothing any of us can do about it. 

 :+1:

Issue #381, opened 2015-04-10
Indent action is inconsistent when paragraph is split with BR tags


DouglasLivingstone:
Using the live demo at http://guardian.github.io/scribe/ in Chrome 41:

1. Type test three times:

    test
    test
    test

Which results in this HTML:

    <p>test</p><p>test</p><p>test</p>

2. Select all the text, then press H2 twice, which results in this HTML:

    <p>test<br>test<br>test</p>

3. Press indent, results in this HTML:

    <blockquote><p>test</p><p>test</p><p>test</p></blockquote>

However, the selection now only extends to the "te" in the final "test", the "st" is unselected.

4. Press Outdent, results in this HTML:

    <p>test</p><p>test</p><p>te</p><blockquote><p><br></p><p>st</p></blockquote>

The "st" which ended up becoming deselected is still indented. Expected result is the html after step 1, with just three paragraph tags. The html at step 2 should probably match the html after step 1 too.

hmgibson23:
If you use the [playground](http://jsbin.com/iwEWUXo/2/edit?js,console,output) to replicate the output in Chrome and Firefox we can better diagnose the issue and see what's what in Scribe.

Thanks for the report.

regiskuckaertz:
@hmgibson23 if this is still occuring I can work on that. can we agree that pressing H2 should in fact produce:

```
<h2>test</h2><h2>test</h2><h2>test</h2>
```

like it would in a page layout/word processor software?

hmgibson23:
@regiskuckaertz - that sounds good to me.

PR #413, opened 2015-06-30
Fixes issue 381

regiskuckaertz:
This fixes the issue https://github.com/guardian/scribe/issues/381 where two characters are left off the selection after the indent command is executed.

This does not fix:
- the behaviour where multiple Ps are automatically converted into a single H2 (step 2 of the process outlined in the issue which is something to do with scribe-plugin-heading-command)
- FF only: the behaviour where the selection is lost and therefore the whole operation fails when `scribe.el` is focused (something to do with scribe-plugin-toolbar)

Not strictly related but I thought could come in handy: the changes in selection serve to spare the need of creating many `Selection` instances by calling the `update` method instead. In fact, I was wondering if `Selection` wouldn't be better as a singleton.


rrees:
Could you add some unit tests to demonstrate the behaviour/fix?

regiskuckaertz:
Sure, I'll work this out today

test/commands.spec.js
```diff
@@ -590,6 +590,20 @@ describe('commands', function () {
...
+    givenContentOf('<p>|test<br>test<br>test|</p>', function () {
+      when('the command is executed', function () {
+        beforeEach(function () {
+          return executeCommand('indent');
+        });
+
+        it('should have no characters left unselected', function () {
+          var selection = new window.scribe.api.Selection();
+          var range = selection.range;
+          return expect(range.endOffset).to.equal(range.endContainer.length);
```
regiskuckaertz: I'm rather new to Chai. Is it how it should be done? I guess we'll see the test results.

rrees:
The tests are failing, if you have the chance to look then I think I'd be happy to merge.

regiskuckaertz:
This may be that I need to bring this PR up to date will the latest changes. I'll look into it when I get back next week. Thanks!

regiskuckaertz:
oh, that is my own test failing. I'll see what's wrong this evening.

regiskuckaertz:
Yeah!

rrees:
Awesome, I'll look into releasing it soon

rrees:
@regiskuckaertz second favour of the day, could you rebase with master so I can release this?

regiskuckaertz:
Sure!

PR #415, opened 2015-07-06
Remove chrome artifacts

regiskuckaertz:
There was some redundancy across several modules regarding an unwanted behaviour in Chrome. I moved the `removeChromeArtifacts` over to the node helpers and rewrote the affected parts. Additionally, I made sure `SPAN`s were removed only if they had no attributes.


src/node.js
```diff
@@ -14,6 +14,16 @@ define([
     return blockElementNames.includes(node.nodeName);
   }
 
+  // Source: https://developer.mozilla.org/en-US/docs/Web/HTML/Inline_elemente
+  var inlineElementNames = Immutable.Set.of('B', 'BIG', 'I', 'SMALL', 'TT',
```
rrees: This might be better in a constants-style module

src/node.js
```diff
@@ -105,7 +115,53 @@ define([
...
+      parentElement,
+      NodeFilter.SHOW_ELEMENT,
+      // arguments to .bind(), starting from the second, are automatically
+      // prepended to the actual arguments when the function is called
+      isInlineWithStyle.bind(null, window.getComputedStyle(parentElement))
+    );
+    var emptySpans = Immutable.List();
+    var node;
+
+    while (node = iterator.nextNode()) {
```
rrees: Could this work as filter? I am assuming not.

src/node.js
```diff
@@ -105,7 +115,53 @@ define([
...
+      node.style.lineHeight = null;
+      if (!node.getAttribute('style')) {
+        node.removeAttribute('style');
+      }
+      if (node.nodeName === 'SPAN' && node.attributes.length === 0) {
+        emptySpans = emptySpans.push(node);
+      }
+    }
+
+    while (!!emptySpans.size) {
```
rrees: You have isEmpty on Immutable.List but it would probably be better as a forEach to make clear it is using side-effects.

I realise this might be less efficient but is clearer.

src/plugins/core/patches/commands/insert-list.js
```diff
@@ -1,4 +1,4 @@
-define([], function () {
```
rrees: There might have been an obscure module bug that meant we kept the empty arrays. Could you submit this change in a different PR so that we don't have yank the consolidation because of the loader.

rrees:
Overall looks sound to me :+1:, would value a second opinion from the team though.

src/node.js
regiskuckaertz: Yes, sure! I assume there will be others with time (like `blockElementNames` for instance), what do you think we put them in a folder, say `/src/vars` or `/src/node/vars`?

src/node.js
regiskuckaertz: Oh yeah, I think it could, something like:

```
var nodes = Immutable.List(parentElement.querySelectorAll(inlineElementNames.join(',')));
nodes = nodes.filter(...);
```

Ideally the selector would select only nodes with a `style` attribute, so this would need to be tweaked. Your call.

src/node.js
regiskuckaertz: :+1:

src/node.js
rrees: Yeah, I think this would be more elegant.

regiskuckaertz:
There it is, indeed it's nicer. Thanks for the suggestion @rrees

rrees:
Small (last) thing could the module be called constants rather than vars. Because they are actually invariants.

regiskuckaertz:
No problemo, I'll take care of this tonight :+1:

PR #416, opened 2015-07-14
Saucelabs: FF 38 and Chrome 43

rrees:
Checking future compatibility with browsers ahead of a future upgrade


cutandpastey:
:+1:

PR #427, opened 2015-09-07
Browser support: September 2015 update

rrees:
The Guardian's official browsers are now: FF40 and Chrome 43/44.

I've also moved the browser support details out of the README and into a wiki page for easy of updating later.


theefer:
Does this need to be reflected [in Composer](https://github.com/guardian/flexible-content/blob/master/composer/src/js/services/helper/supported-browser-collection.js) ?

rrees:
Yes it does, FF40 is being rolled out now (visible in Mixpanel) so next week might be good to make the change in Composer

Issue #425, opened 2015-09-03
Determine keydown event when execute is called


josh-infusionsoft:
I am creating a plugin and there is a certain case where execute is called by either [enter or backspace] (https://github.com/guardian/scribe/blob/master/src/plugins/core/events.js#L163) in a list and I need to know which key it was.

Is there a known way to get this information (must work in FF and chrome)? Or will I need to modify scribe in order to accomplish this?

josh-infusionsoft:
Here is my attempt at a fix. If you have a better suggestion I would love to hear it:

https://github.com/guardian/scribe/pull/428

rrees:
Please test release v2.0.1 and close if fixed. Thanks for the PR.

PR #428, opened 2015-09-08
Add event to command

josh-infusionsoft:
While navigating an ordered list, execute is called when a backspace or enter key is pressed. A plugin might want to know which key was pressed so it can decide how to handle the request. Event has been added to the command.

### DIFF
--- a/src/plugins/core/events.js
+++ b/src/plugins/core/events.js
@@ -160,6 +160,8 @@ define([
 
                 var command = scribe.getCommand(listNode.nodeName === 'OL' ? 'insertOrderedList' : 'insertUnorderedList');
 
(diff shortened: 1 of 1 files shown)


josh-infusionsoft:
Here is the issue associated with this pull request:
https://github.com/guardian/scribe/issues/425

rrees:
This seems reasonable to me, although it might be nice to be more consistent about passing an event to the Command in future.

josh-infusionsoft:
Would you like me to change it? Or are you willing to merge as is?

rrees:
@josh-infusionsoft I'll merge as is but if you have time to propose another PR that'll be great.

Issue #433, opened 2015-09-28
Paste does not work on android


crasu:
Pasting text into scribe does not work for me on android 5.x. 

The mobile chrome paste event does not contain the pasted content in event.clipboardData.types. However the code around line 188 in events.js seems to expect this. Adding a check event.clipboardData.types seems to fix the issue. 

Would you merge a pull request for this issue even if mobile chrome is not officially supported?

rrees:
@crasu happy to take a pull request but obviously as it would not be in the test suite it might be vulnerable to regressions. Add a comment on the check to make clear why it is there.

crasu:
see #434

PR #434, opened 2015-09-29
Fix paste on android

crasu:
see issue #433


rrees:
Thanks for the fix, sorry it took a while to get out, SauceLabs tests were being flakey.

PR #436, opened 2015-10-15
Add skippable node in MutationObserver

SimonDegraeve:
I mount [React](https://facebook.github.io/react/) component inside the Scribe container, whenever the component update it triggers the MutationObserver watching the subtree of the Scribe container. 
Since the component is already encapsulate in the React tree, I don't need this extra re-render from Scribe.


rrees:
@SimonDegraeve can you check that Scribe 2.1.0 implements this change correctly?

PR #438, opened 2015-10-21
Enable server side rendering

szyablitsky:
This patch enables server side rendering, allowing `window` access at runtime only if `window` defined.

PR #440, opened 2015-11-03
Server-side Scribe rendering support

rrees:
Combines #438 and #436 with a unit-testing strategy that might help us get around the flakiness in the browser driven tests.

@szyablitsky  @SimonDegraeve can you check that this code matches your intent?

### DIFF
--- a/src/dom-observer.js
+++ b/src/dom-observer.js
@@ -1,10 +1,9 @@
 define([
-  './node'
-], function (nodeHelpers) {
+  './node',
+  './mutations'
+], function (nodeHelpers, mutations) {
 
-  var MutationObserver = window.MutationObserver ||
-    window.WebKitMutationObserver ||
-    window.MozMutationObserver;
+  var MutationObserver = mutations.determineMutationObserver(window);
 
   function hasRealMutation(n) {
     return ! nodeHelpers.isEmptyTextNode(n) &&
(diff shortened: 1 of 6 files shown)


SimonDegraeve:
It sounds good to me.

szyablitsky:
LGTM :+1:

src/mutations.js
```diff
@@ -0,0 +1,22 @@
+define([], function() {
+
+  function determineMutationObserver(window) {
```
szyablitsky: @rrees Server side rendering still does not work.
You pass the `window` to the function. And `window` is not defined on server.
I guess you added this parameter only for testing purposes. But it nullifies the intent of this patch.
May be you can mock window object inside the test? So we do not need to pass it as a parameter.

rrees: No this one is meant to be a parameter, I'll look at what we can do in the caller

src/dom-observer.js
```diff
@@ -1,10 +1,9 @@
...
-  './node'
-], function (nodeHelpers) {
+  './node',
+  './mutations'
+], function (nodeHelpers, mutations) {
 
-  var MutationObserver = window.MutationObserver ||
-    window.WebKitMutationObserver ||
-    window.MozMutationObserver;
+  var MutationObserver = mutations.determineMutationObserver(window);
```
szyablitsky: @rrees here you go. Calling the function with `window` as a param.

PR #441, opened 2015-11-12
Server-side rendering: default value for window

rrees:
When the window global is undefined (in a Node environment) define it to be undefined.

Ping @szyablitsky

### DIFF
--- a/src/dom-observer.js
+++ b/src/dom-observer.js
@@ -3,7 +3,9 @@ define([
   './mutations'
 ], function (nodeHelpers, mutations) {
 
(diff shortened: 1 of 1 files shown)


szyablitsky:
I tried this solution locally in node_modules folder. And it works.
Thank you. Let's merge it :+1:

szyablitsky:
@rrees I updated package and enabled server side rendering again. All is working like a charm. Thank you again! :satisfied:

rrees:
@szyablitsky Great news!

PR #443, opened 2015-11-16
Formatters that returns Promises

triptec:
This is just how I could work I guess, thoughts?

### DIFF
--- a/src/plugins/core/events.js
+++ b/src/plugins/core/events.js
@@ -55,8 +55,10 @@ define([
             if (isEditorActive) {
               selection.placeMarkers();
             }
-            scribe.setHTML(scribe._htmlFormatterFactory.format(scribe.getHTML()));
-            selection.selectMarkers();
+            scribe._htmlFormatterFactory.format(scribe.getHTML()).then(function(html){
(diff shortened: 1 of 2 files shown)


rrees:
With the state of the test suite this is something that might take some time to get this merged. This is to allow user-functions to be chained to the formatters right?

theefer:
Nice idea, but there may be wide ranging implications with making formatters async that will require thorough testing to avoid breaking current code.

triptec:
Well my use case is that if there is a link pasted in the editor I would like to substitute it with a [link src='' type=internal | external etc] and to be able to determine if it's a internal src I need to make a http get and therefor return a promise.

I guess most plugins won't need to return a promise and that works to, obviously we need to get the tests running, I might be able to help with that, just wanted to know if this is something you could see going in to the codebase.

I also would like to know if there's a better way to handle my use case but I still think it would be nice to have formatters able to handle promises if needed.

This might have a performance impact though and that could be a good reason not to make it work this way, I guess we could add async formatters instead of changing a current api.

rrees:
Yeah, adding a specific async formatter is definitely going to be easier to get into production

PR #453, opened 2016-02-27
Undo: allow undo to be switched off

### LINKED ISSUE #452: Undo manager cannot be disabled
If I set undo to disabled, scribe complains because _lastItem is null and part of the constructor calls setHTML which has a call for

`_lastItem.content = <whatever>`

There needs to be an else block to allow for this. Currently, my work around is to simply set

`Scribe.prototype._lastItem = {content: ''};`

rrees:
Fixes #452 by simply putting an if condition around the history interaction in setHTML.

Adds an example page that disables the undo manager.

### DIFF
--- a/examples/no-undo-manager.html
+++ b/examples/no-undo-manager.html
@@ -1,0 +1,105 @@
+<!--
+This example demonstrates how to consume the Scribe
+editor using RequireJS and the AMD module syntax.
+
+Note that you'll need to install scribe's dependencies through
+`bower install`. See http://bower.io/ if you are unfamiliar.
+-->
+<style>
+  button {
+    height: 3em;
+  }
+
+  .active {
+    border-style: inset;
+  }
+
+  .rte, .rte-toolbar {
+    display: block;
+  }
+
+  p {
+    margin-top: 0;
+  }
+
+  .rte {
+    border: 1px solid gray;
+    height: 300px;
+    overflow: auto;
+  }
+  .rte-output {
+    width: 100%;
+    height: 10em;
+  }
+</style>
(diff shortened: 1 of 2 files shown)


src/scribe.js
```diff
@@ -133,7 +133,9 @@ define([
   };
 
   Scribe.prototype.setHTML = function (html, skipFormatters) {
-    this._lastItem.content = html;
+    if (this.options.undo.enabled) {
```
nedrocks: I believe this works but since I'm not super well versed I'd be more tempted to do something closer to:

`this._lastItem.content = {content: ''}`

This works for me and I have it deployed to (well okay, Beta) production without any issues. Not sure if there are any other weird codepaths that we haven't found yet where _lastItem is depended on without a check somewhere.

examples/no-undo-manager.html
```diff
@@ -0,0 +1,105 @@
+<!--
```
nedrocks: I don't know enough about AMD to comment very intelligently on this file but it looks good to me.

nedrocks:
Bunch of unit test failures but they don't appear to be related to this change -- are these known failures?

rrees:
@nedrocks Yeah, unfortunately it's a long story. See #454 for an unsatisfactory fix for the unit tests

src/scribe.js
rrees: That approach breaks the undo functionality though.

nedrocks: Ah good point! Okay, then point withdrawn and LGTM!

nedrocks:
LGTM!

PR #479, opened 2016-04-29
Immutable: update dependency

rrees:
This is a simple keeping us current update that moves to the latest minor release version.


hmgibson23:
👍

Issue #282, opened 2014-10-30
Undo manager prevents entering Alt + Z special characters


pwiszowaty:
Hey guys, 

we were just reported that some of our users can't enter special chars like ż, Ż (alt + z). 
That event.metaKey in undo / redo commands messes it up. Can't think of any good way to fix this, so I'm not creating a pull request. 

In our case - we just stopped using undo / redo by removing the key bindings, but it would be great to be able to use those while not loosing those special Z chars.

paperboyo:
Related: https://github.com/guardian/scribe/issues/448

danburzo:
Should be fixed now!

PR #480, opened 2016-05-01
Fixes #448, #282; allow entering polish ż

### LINKED ISSUE #448: Can't type polish letter ż with "Polish Pro" keyboard layout on Windows
In short, `Right Alt` + `Z` on a Polish Pro keyboard layout will trigger an event with:

```js
{
  ctrlKey: true,
  altKey: true,
  ...
}
```

which is caught by the undo manager. 

[This article from Medium Engineering](https://medium.com/medium-eng/the-curious-case-of-disappearing-polish-s-fa398313d4df#.2n55jevdw) goes into more details on why this happens.

~~I've forked Scribe and will be providing a Pull Request with the fix.~~ Can't currently install Scribe tests because of this: https://github.com/guardian/scribe-test-harness/issues/26

danburzo:
Now that I seem to be able to run Scribe locally, tests & all, I'm providing the fix I proposed on #448.
I'm not sure how to approach writing a test for this, any ideas?

### DIFF
--- a/src/keystrokes.js
+++ b/src/keystrokes.js
@@ -1,0 +1,17 @@
+define(function() {
+
+  'use strict';
+
+  function isUndoKeyCombination(event) {
+    return !event.shiftKey && (event.metaKey || (event.ctrlKey && !event.altKey)) && event.keyCode === 90;
+  }
+
+  function isRedoKeyCombination(event) {
+    return event.shiftKey && (event.metaKey || (event.ctrlKey && !event.altKey)) && event.keyCode === 90;
+  }
+
+  return {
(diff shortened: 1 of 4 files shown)


rrees:
Thanks @danburzo. My best idea for testing this is to take the very complex test in the if and put in its own testable function, perhaps called something like `isRedoKeyCombination`, and then write a test against that.

We don't have a module for this at the moment so maybe create a top-level module, something like `keystrokes.js`?

danburzo:
Hi @rrees! I've just added a new `keystrokes` module and some tests.

rrees:
Looks super good to me. :+1: 

I'll try and organise a release as soon as I have time.

PR #484, opened 2016-06-13
Expanded the Options section, added caveat in Production-Ready section

danburzo:
I propose two changes to README.md:

1. Better documentation of user-configurable options, which are not apparent (as far as I could tell) unless reading through the source code. 
2. A caveat in the _production-ready_ section, since it was a bit surprising that after a (long-due) update in our Scribe instance, it stopped working in Safari 8 and Chrome 45 (due to missing support for `Object.assign` -- which BTW is easily patched with a polyfill, something worth considering?)

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,14 +52,91 @@ Alternatively, you can [access the distribution files through GitHub releases](h
 
 ## Options
 
-<dl>
-  <dt><pre>allowBlockElements</pre></dt>
-  <dd>Enable/disable block element mode (enabled by default)</dd>
-  <dt><pre>defaultCommandPatches</pre></dt>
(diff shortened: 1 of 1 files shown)


rrees:
I can see why you want the expanded documentation but I might add them into the wiki or a docs folder rather than README since they are not things you need to access frequently. I will also add a version note to things like the default plugin list which will change in future. Hope that's okay with you.

Version 3 of Scribe was a deliberately breaking change (hence a major version number) but you're right that it would be sensible to add some notes about what people might need to polyfill for older browsers.

Thanks for contributing!

danburzo:
Sounds good! 👍

PR #490, opened 2016-07-22
Fixes #489

### LINKED ISSUE #489: Pressing Enter while having a selection in inline-mode does not delete the range content
The code that handles inline mode (`allowBlockElements=false`) assumes that when pressing the Enter key, the range is always collapsed. When it happens not to be, the selected characters are not deleted when inserting the new line.

danburzo:
This pull request fixes some issues identified in #489 for inline-mode:

**Pressing Enter while having a selection** originally did not delete the range contents. It now does.

**Pressing Enter in the middle of a word** would insert two `<br>` elements between the parts. I've changed the `hasContent` method such that it takes into consideration text nodes that have non-whitespace content. (created a new nodeHelpers method + related tests).

I've also added a `raw-inline.html` example for fiddling with inline-mode Scribe.


danburzo:
I've also included a sanity check for `scribe.el.lastChild`. Scribe assumes the element always has some child, but there are situations when the element gets completely empty. Pressing Enter in these situations will throw an error.

rrees:
Looks good 👍

codeclown:
Just noticed this too. When would this be merged?

PR #492, opened 2016-07-22
Fixes #488; allow new line at the end of the line in Safari

### LINKED ISSUE #488: allowBlockElements=false in Safari prevents entering new lines
If you instantiate Scribe with `allowBlockElements=false` in Safari, you're not able to insert new lines with the Enter key. I'm investigating the source of the bug and will propose a PR, but any guidance in the meantime is appreciated!

danburzo:
Fixes the bug where you could not insert a new line at the end of the line in Safari, when using inline-mode.


danburzo:
My pleasure, glad to have you on the project!

PR #501, opened 2016-10-22
Do not insert `<br>` into custom elements

nazar-pc:
I only found tests for formatter in `test-old` but don't know how to run them (they are not running with `npm test`) so no tests this time.

I've started using Scribe with Shadow DOM and custom elements, so expect more PRs in future if there will be any issues:)


nazar-pc:
I don't have an example live or in public repository, but the project I'm working on will be 100% open source eventually (hopefully soon). Maintaining forks is pretty difficult from my experience, so I'm submitting upstream all of the patches needed so far to make Scribe compatible with custom elements (including Shadow DOM).

PR #503, opened 2016-10-26
Added Shadow DOM awareness to selection

nazar-pc:
We need to use closest Shadow Root instead of Document for elements withing Shadow Root.
Otherwise we'll get no useful information about actual selection.

### DIFF
--- a/src/api/selection.js
+++ b/src/api/selection.js
@@ -9,8 +9,21 @@ define(function () {
     // find the parent document or document fragment
     if( rootDoc.compareDocumentPosition(scribe.el) & Node.DOCUMENT_POSITION_DISCONNECTED ) {
       var currentElement = scribe.el.parentNode;
-      while(currentElement && nodeHelpers.isFragment(currentElement)) {
-        currentElement = currentElement.parentNode;
+      // Old browsers might not know about `:host` without polyfill, lets catch exception if it is the case
+      try {
+        var withinShadowRoot = currentElement.matches && currentElement.matches(':host *');
+      } catch (e) {}
+
+      // We need to use closest Shadow Root instead of Document for elements withing Shadow Root
+      // Otherwise we'll get no useful information about actual selection
+      if (withinShadowRoot) {
+        while (!currentElement.host) {
+          currentElement = currentElement.parentNode;
+        }
+      } else {
+        while(currentElement && nodeHelpers.isFragment(currentElement)) {
+          currentElement = currentElement.parentNode;
+        }
       }
 
       // if we found a document fragment and it has a getSelection method, set it to the root doc


nazar-pc:
Somebody alive here?

src/api/selection.js
```diff
@@ -9,8 +9,21 @@ define(function () {
     // find the parent document or document fragment
     if( rootDoc.compareDocumentPosition(scribe.el) & Node.DOCUMENT_POSITION_DISCONNECTED ) {
       var currentElement = scribe.el.parentNode;
-      while(currentElement && nodeHelpers.isFragment(currentElement)) {
-        currentElement = currentElement.parentNode;
+      // Old browsers might not know about `:host` without polyfill, lets catch exception if it is the case
```
rrees: I think this check should be in a different module

src/api/selection.js
```diff
@@ -9,8 +9,21 @@ define(function () {
...
+      try {
+        var withinShadowRoot = currentElement.matches && currentElement.matches(':host *');
+      } catch (e) {}
+
+      // We need to use closest Shadow Root instead of Document for elements withing Shadow Root
+      // Otherwise we'll get no useful information about actual selection
```
rrees: The conditional nesting is gnarly here. Have the while loop take a function and then use the `withinShadowRoot` flag to choose which terminating function to use.

rrees (commented):
In some ways it feels like you should put Scribe into a ShadowDom mode in the options and then condition the behaviour off that.

Apart from comments though, looks sensible.

src/api/selection.js
nazar-pc: Do you mean AMD module here or what?

src/api/selection.js
nazar-pc: I agree, but don't quite understand how do you want it to be changed.
BTW, feel free to modify it if necessary (maintainers should be able to do that directly, otherwise send a PR against my fork).

src/api/selection.js
rrees: Yes, exactly. So this is the selection module so it feels like the wrong place for a Shadow DOM check, putting it in a Shadow DOM module and then importing it would make it easier to maintain.

src/api/selection.js
rrees: I've actually left the Guardian so I no longer have admin rights to the repo. This is kind of a drive-by reviewing.

I think the code should roughly be:

```
var/const/let rootCheck = isShadowDom(currentElement) ? shadowRootCheck : domRootCheck;

while(!rootCheck(currentElement)) {
  currentElement = currentElement.parentNode;
}
```

Make sense?

PR #506, opened 2017-01-30
fix npm installation

### LINKED ISSUE #495: lodash dependecy still exists.
The latest scribe-editor installed through npm results in dependency errors.
It says cannot resolve module 'lodash-amd/modern/string/escape' in the below file
https://github.com/guardian/scribe/blob/master/src/plugins/core/formatters/plain-text/escape-html-characters.js

lodash-amd is not listed as dependency,so relevant files are not downloaded.
I had to add lodash-amd@3.5.0 as dependency to resolve this,the latest lodash-amd doesn't maintain same folder structure, so it's of no help.

This error doesn't not happen when scribe is downloaded from bower. The reason being there are polyfills for that particular dependency thats being added to main scribe.js source file.

I propose adding same polyfill to the file mentioned above to resolve this issue instead of adding and entire obsolete library.
Will make an PR and reference this issue.
Thanks

ShaunYearStrong:
This adds `lodash-npm@3.5.0` as a dependency in `package.json` to match the bower installation.

This will fix #495, but I made the decision to use the obsolete library instead of the polyfill.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,8 @@
   "license": "Apache-2.0",
   "main": "src/scribe.js",
   "dependencies": {
-    "immutable": "~3.8.1"
+    "immutable": "~3.8.1",
+    "lodash-amd": "3.5.0"
   },
   "devDependencies": {
     "chai": "~1.9.1",


code-smith:
@ShaunYearStrong  This is not the right way to go ahead.
#1 Lots of code changes in previous releases were aimed at making library independent of lodash, this will just undo everything.
#2 lodash-amd@3.5.0 is older version, most use lodash-4 , its very different in code structure to 3.5. Any user application will end up having both lodash-4 and lodash-3.5 which can be avoided if the polyfill is accepted

It's your decision, but personally i don't think its the right way to go ahead.It bloats up every application code which uses the latest version of lodash. I think you should give it another thought.

Thanks anyways.

PR #511, opened 2017-06-29
add save and resotre function to Selection

xsbchen:


shaundillon:
Hi @xsbchen Thanks for your contribution.
Could you give some details about the use cases here?
I feel that, even though it's dealing with a selection, the selection module is the wrong place to put this. It's more additional logic than just selection logic.
Perhaps a plugin would be more suitable?

PR #522, opened 2018-08-09
Fix Scribe adding links to other contenteditables

RichieAHB:
This PR stops Scribe throwing it's links around into other contenteditable nodes. This probably could be widened to ignore any commands when Scribe doesn't have a selection or focus, but the surface area of this PR would increase by a factor of 100 and I have no idea what it might break.

### DIFF
--- a/src/api/selection.js
+++ b/src/api/selection.js
@@ -101,35 +101,36 @@ define(function () {
         nodeHelpers.getAncestor(node, scribe.el, nodeFilter);
     };
 
-    Selection.prototype.placeMarkers = function () {
+    Selection.prototype.isInScribe = function () {
       var range = this.range;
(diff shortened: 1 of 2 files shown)


src/api/selection.js
```diff
@@ -101,35 +101,36 @@ define(function () {
         nodeHelpers.getAncestor(node, scribe.el, nodeFilter);
     };
 
-    Selection.prototype.placeMarkers = function () {
+    Selection.prototype.isInScribe = function () {
```
RichieAHB: The diff doesn't really show this nicely but `placeMarkers` still exists but it has functionality to check whether the selection was in Scribe that I broke out into a method that could be used in `create-link`.

PR #529, opened 2019-01-30
Add deprecation information

RichieAHB:
Add the deprecation information to the README and link to [Leaving Scribe](https://www.theguardian.com/info/2019/jan/24/leaving-scribe). I will also prefix the GitHub description with `DEPRECATED: ` once this is merged.

Any comments welcome 👍

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,9 @@
+**THIS PROJECT IS DEPRECATED** - You can find more information about this in our blog post, [Leaving Scribe](https://www.theguardian.com/info/2019/jan/24/leaving-scribe). In summary:
+
+- We have no plans to add features to Scribe but may make critical updates throughout the period that we continue to use instances of Scribe internally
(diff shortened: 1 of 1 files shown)


RichieAHB:
Also I'm not sure what happened to `html-janitor` I think we recovered it after having deleted it but it's failing the build as we maybe didn't recover it to the version it was at before deletion (notable `2.0.4`). I can either downgrade Scribe back to `2.0.2` but I feel like the patch versions in `2.0.4` probably fixed something worth fixing!

mbarton:
@RichieAHB I published 2.0.4 a while back to fix some vulnerabilities: https://github.com/guardian/html-janitor/pull/1

It shows as the latest version on npm (https://www.npmjs.com/package/html-janitor). It was 2.0.2 before deletion from memory, with 2.0.4 being my version I released after deleting the repo.

RichieAHB:
@mbarton ah interesting - the `package.json` shows as `2.0.4` so it's probably just that some tags got deleted - happy to push 2.0.4 to master.

